Issue #1, opened 2017-09-20
Current db_parameter_group module doesn't work with Postgres


khosrow:
The parameters listed here in [Line 24](https://github.com/terraform-aws-modules/terraform-aws-rds/blob/0750cc1c749efc95a3ac510525a73965d268c930/modules/db_parameter_group/main.tf#L24) are MySQL specific.

antonbabenko:
Fixed by #2

PR #2, opened 2017-09-20
db parameters can now be passed to db_parameter_group module

khosrow:
Addresses #1 as well.

**Note:** Was tested successfully with Postgres engine in RDS.


antonbabenko:
Thanks for the contribution! You are the first who participated since the official release of the Terraform registry 👍

PR #3, opened 2017-09-21
support final_snapshot_identifier

jensendw:
In its current state the module will provision an RDS instance just fine, however, because skip_final_snapshot is set to false by default AWS wont allow the RDS instance to be deleted.  The only way I found to solve this was to require the setting of final_snapshot_identifier as I wanted to be conservative and not default to skipping the final snapshot by default.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,39 +23,42 @@ module "db" {
   source = "terraform-aws-modules/rds/aws"
 
   identifier = "demodb"
-  
+
   engine            = "mysql"
   engine_version    = "5.7.11"
   instance_class    = "db.t2.large"
   allocated_storage = 5
-  
+
   name     = "demodb"
   username = "user"
   password = "YourPwdShouldBeLongAndSecure!"
   port     = "3306"
-  
+
   vpc_security_group_ids = ["sg-12345678"]
-  
+
   maintenance_window = "Mon:00:00-Mon:03:00"
   backup_window      = "03:00-06:00"
-  
+
   tags = {
     Owner       = "user"
     Environment = "dev"
   }
-  
+
   # DB subnet group
   subnet_ids = ["subnet-12345678", "subnet-87654321"]
-  
+
   # DB parameter group
   family = "mysql5.7"
 
+  # Snapshot name upon DB deletion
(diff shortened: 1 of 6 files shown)


modules/db_instance/variables.tf
```diff
@@ -39,6 +39,10 @@ variable "port" {
   description = "The port on which the DB accepts connections"
 }
 
+variable "final_snapshot_identifier" {
+  description = "The name of your final DB snapshot when this DB instance is deleted."
+}
```
antonbabenko: There should be `default = ""`, because this is an optional argument.

variables.tf
```diff
@@ -19,6 +19,10 @@ variable "engine_version" {
   description = "The engine version to use"
 }
 
+variable "final_snapshot_identifier" {
+  description = "The name of your final DB snapshot when this DB instance is deleted."
+}
```
antonbabenko: Same here.

README.md
```diff
@@ -23,39 +23,42 @@ module "db" {
   source = "terraform-aws-modules/rds/aws"
 
   identifier = "demodb"
-  
+
```
antonbabenko: Good catch with all these spaces.

modules/db_instance/variables.tf
jensendw: Since skip_final_snapshot defaults to false (https://github.com/terraform-aws-modules/terraform-aws-rds/blob/master/modules/db_instance/variables.tf#L108) And the AWS api wont allow you to delete an RDS instance without a final_snapshot_identifier set unless skip_final_snapshot = true, I'm not sure what is the best setting here.  It seems to be that either skip_final_snapshot should be true by default, which I would consider a bad design decision as it could have unwanted effects.  So my thought was to force the identification of final_snapshot_identifier instead.

antonbabenko: This resource module does not have to be smarter than Terraform & AWS API, so keeping it as `default = ""` is a good solution. Please fix and we merge this PR.

jensendw: 👍

PR #4, opened 2017-09-27
Using `storage_encrypted` to set the db_instance.storage_encrypted

fmartingr:
I've added `encrypted` as a module variable to passthrough the value to the `aws_db_instance`.`storage_encrypted` parameter.

The documentation does not specify that more parameters are required but I'm not 100% sure to be honest.

### DIFF
--- a/examples/complete/main.tf
+++ b/examples/complete/main.tf
@@ -30,6 +30,8 @@ module "db" {
   engine_version    = "5.7.11"
   instance_class    = "db.t2.large"
   allocated_storage = 5
+  storage_encrypted = true
+  # kms_key_id        = "arm:aws:kms:<region>:<accound id>:key/<kms key id>"
 
   name     = "demodb"
   username = "user"
--- a/examples/complete/main.tf
+++ b/examples/complete/main.tf
@@ -30,7 +30,7 @@ module "db" {
   engine_version    = "5.7.11"
   instance_class    = "db.t2.large"
   allocated_storage = 5
-  storage_encrypted = true
+  storage_encrypted = false
   # kms_key_id        = "arm:aws:kms:<region>:<accound id>:key/<kms key id>"
 
   name     = "demodb"
--- a/main.tf
+++ b/main.tf
@@ -65,7 +65,5 @@ module "db_instance" {
   backup_retention_period = "${var.backup_retention_period}"
   backup_window           = "${var.backup_window}"
 
-  storage_encrypted = "${var.encrypted}"
-
   tags = "${var.tags}"
 }
--- a/main.tf
+++ b/main.tf
@@ -39,6 +39,7 @@ module "db_instance" {
   instance_class    = "${var.instance_class}"
   allocated_storage = "${var.allocated_storage}"
   storage_type      = "${var.storage_type}"
+  storage_encrypted = "${var.storage_encrypted}"
 
   name     = "${var.name}"
   username = "${var.username}"
--- a/main.tf
+++ b/main.tf
@@ -40,6 +40,7 @@ module "db_instance" {
   allocated_storage = "${var.allocated_storage}"
   storage_type      = "${var.storage_type}"
   storage_encrypted = "${var.storage_encrypted}"
+  kms_key_id        = "${var.kms_key_id}"
 
   name     = "${var.name}"
   username = "${var.username}"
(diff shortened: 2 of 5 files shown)


main.tf
```diff
@@ -63,5 +63,7 @@ module "db_instance" {
   backup_retention_period = "${var.backup_retention_period}"
   backup_window           = "${var.backup_window}"
 
+  storage_encrypted = "${var.encrypted}"
```
antonbabenko: Keep variables names exactly as in Terraform internally. `storage_encrypted` in this case.

antonbabenko: Also move it after `storage_type`, as these two parameters are related to storage.

modules/db_instance/variables.tf
```diff
@@ -116,3 +116,8 @@ variable "tags" {
   description = "A mapping of tags to assign to all resources"
   default     = {}
 }
+
+variable "storage_encrypted" {
+  description = "Defines if the storage should be encrypted or not"
```
antonbabenko: Keep description as in official documentation - `Specifies whether the DB instance is encrypted.`. See here - https://www.terraform.io/docs/providers/aws/r/db_instance.html

variables.tf
```diff
@@ -112,6 +112,11 @@ variable "tags" {
   default     = {}
 }
 
+variable "encrypted" {
+  description = "Defines wether this instance should be encrypted or not"
```
antonbabenko: Same here

modules/db_instance/variables.tf
antonbabenko: And also move it after `storage_type`.

fmartingr:
@antonbabenko So, I've been making some tests and changed what you've asked.

Regarding the `kms_key_id`: AWS seems to create one automatically if you don't define the parameter. I didn't have one and creating an instance without it created one on my account, so I allowed a default of `''` (empty string) there but added the parameters to the variables anyway.

modules/db_instance/variables.tf
```diff
@@ -11,6 +11,16 @@ variable "storage_type" {
   default     = "gp2"
 }
 
+variable "storage_encrypted" {
+  description = "Specifies whether the DB instance is encrypted"
+  default     = false
+}
+
+variable "kms_key_id" {
+  description = "The ARN for the KMS encryption key. If creating an encrypted replica, set this to the destination KMS ARN."
```
antonbabenko: Add a note as you discovered like: `If storage_encrypted is set to true and kms_key_id is not specified the default KMS key created in your account will be used`.

modules/db_instance/main.tf
```diff
@@ -34,5 +36,7 @@ resource "aws_db_instance" "this" {
   backup_retention_period = "${var.backup_retention_period}"
   backup_window           = "${var.backup_window}"
 
+  storage_encrypted = "${var.storage_encrypted}"
```
antonbabenko: duplicated

variables.tf
```diff
@@ -11,6 +11,16 @@ variable "storage_type" {
   default     = "gp2"
 }
 
+variable "storage_encrypted" {
+  description = "Specifies whether the DB instance is encrypted"
+  default     = false
+}
+
+variable "kms_key_id" {
+  description = "The ARN for the KMS encryption key. If creating an encrypted replica, set this to the destination KMS ARN."
```
antonbabenko: Same here.

antonbabenko:
Thanks. Yes, this is typical AWS behavior.

fmartingr:
Wooops, sorry @antonbabenko, I pushed with VSCode and it messed up my commits, there was also a duplicated variable there. Everything seems is fine now I think 😓

variables.tf
antonbabenko: This variable is not needed.

examples/complete/main.tf
```diff
@@ -30,6 +30,8 @@ module "db" {
   engine_version    = "5.7.11"
   instance_class    = "db.t2.large"
   allocated_storage = 5
+  storage_encrypted = true
```
antonbabenko: Let's turn encryption off in this example because it will create KMS key which we don't manage anyhow and it won't be destroyed by `terraform destroy`.

fmartingr: Yeah, so true.

PR #6, opened 2017-09-30
Add enhanced monitoring

robinbowes:
Re-working of https://github.com/terraform-community-modules/tf_aws_rds/pull/37

(plus a `terraform fmt` :) )

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,8 +40,8 @@ module "db" {
   backup_window      = "03:00-06:00"
 
   # Enhanced Monitoring - see example for details on how to create the role
-  # monitoring_interval = "10"
-  # monitoring_role_arn = "aws_iam_role.rds_enhanced_monitoring.arn"
+  monitoring_interval = "30"
+  monitoring_role_arn = "arn:aws:iam::123456789012:role/rds-monitoring-role"
   
   tags = {
     Owner       = "user"
--- a/README.md
+++ b/README.md
@@ -38,6 +38,10 @@ module "db" {
   
   maintenance_window = "Mon:00:00-Mon:03:00"
   backup_window      = "03:00-06:00"
+
+  # Enhanced Monitoring
+  monitoring_interval = "${var.monitoring_interval}
+  monitoring_role_arn = "${var.monitoring_interval == "0" ? "" : aws_iam_role.rds_monitoring.arn}"
   
   tags = {
     Owner       = "user"
--- a/README.md
+++ b/README.md
@@ -39,9 +39,9 @@ module "db" {
   maintenance_window = "Mon:00:00-Mon:03:00"
   backup_window      = "03:00-06:00"
 
-  # Enhanced Monitoring
-  monitoring_interval = "${var.monitoring_interval}
-  monitoring_role_arn = "${var.monitoring_interval == "0" ? "" : aws_iam_role.rds_monitoring.arn}"
+  # Enhanced Monitoring - see example for details on how to create the role
+  # monitoring_interval = "10"
+  # monitoring_role_arn = "aws_iam_role.rds_enhanced_monitoring.arn"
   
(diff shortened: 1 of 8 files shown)


examples/enhanced_monitoring/README.md
```diff
@@ -0,0 +1,5 @@
+Enhanced Monitoring example
+===========================
+
+Configuration in this directory creates the additional resources required to use Enhanced Monitoring.
```
antonbabenko: Remember that these examples should always be executed using just terraform init, plan, apply. Users should not be asked to update anything in the example to try them out. Please include "provider" section and "rest of params here as per complete example".

README.md
```diff
@@ -38,6 +38,10 @@ module "db" {
   
   maintenance_window = "Mon:00:00-Mon:03:00"
   backup_window      = "03:00-06:00"
+
+  # Enhanced Monitoring
+  monitoring_interval = "${var.monitoring_interval}
+  monitoring_role_arn = "${var.monitoring_interval == "0" ? "" : aws_iam_role.rds_monitoring.arn}"
```
antonbabenko: Similar should be in usage section. Most users want to just copy-paste this from readme file and put as little effort as possible to get it up, but line 44 will make them think that they need to create IAM role somehow.

Keep this section as easy as possible and make it look as end result:
```
monitoring_interval = "10"
monitoring_role_arn = "arn:aws:i-don't-know-but-you-know-what-to-write-here:12345678"
```

No need to use variables here - again, for simplicity.

examples/enhanced_monitoring/README.md
robinbowes: Are there guidelines for examples?

antonbabenko: No, but there will be (probably during next week or two).

README.md
robinbowes: Yeah, I didn't mean to include that.

But, there's no magic role arn you can include - you have to create the role. Not sure how best to document it.

antonbabenko: I think it will be best to make an example, where you create everything what is needed for it. And link to this example on the main readme page.

Maybe soon we can also start section "Features", where we can highlight what this module can do. Something like:
```
Features:
* Create RDS resources with most of the options supported
* Support [enhanced monitoring](example/...). This module does not create IAM role for you. Check example code for more details.
```

robinbowes: OK, I'll pull the params from the README and fix pu the example.

robinbowes:
How's this?

README.md
```diff
@@ -39,9 +39,9 @@ module "db" {
   maintenance_window = "Mon:00:00-Mon:03:00"
   backup_window      = "03:00-06:00"
 
-  # Enhanced Monitoring
-  monitoring_interval = "${var.monitoring_interval}
-  monitoring_role_arn = "${var.monitoring_interval == "0" ? "" : aws_iam_role.rds_monitoring.arn}"
+  # Enhanced Monitoring - see example for details on how to create the role
+  # monitoring_interval = "10"
+  # monitoring_role_arn = "aws_iam_role.rds_enhanced_monitoring.arn"
```
antonbabenko: Write how the end result will have to look like - `monitoring_role_arn = "arn:aws:iam::123456789012:role/RDSEnhancedMonitoring"` and leave it to curious user to figure out how to get the value (maybe he will use data-source to find the one he has already make, maybe he will create resource himself, maybe he will hardcode it, maybe he passes it as variable, ...)

robinbowes: Is it worth putting in a link to the relevant AWS page?

antonbabenko: I think so, a small note is usually a good idea. In general, I would love to hear from real users what do they miss.

robinbowes:
Did you get chance to look at this?

PR #11, opened 2017-10-11
Use relative links

robinbowes:
Switch to use relative links where possible in the README

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,9 +11,9 @@ These types of resources are supported:
 
 Root module calls these modules which can also be used separately to create independent resources:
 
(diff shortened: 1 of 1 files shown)


antonbabenko:
This is a bad idea for the Terraform Registry.

robinbowes:
Does it use the same README? It doesn't appear to have updated since you tagged the new version?

antonbabenko:
Yes, it does if you follow the default link - https://registry.terraform.io/modules/terraform-aws-modules/rds/aws/ or [specific version](https://registry.terraform.io/modules/terraform-aws-modules/rds/aws/1.0.5)

robinbowes:
Ah, I see - I had assumed my README changes were in the 1.0.5 release - they are not.

OK, I'll close this.

R.

PR #19, opened 2017-10-25
Fixed replica support

kwent:
db_subnet_group_name must *NOT* be specified when creating aws_rds_cluster_instance replica resource


kwent:
Nevermind this needs much more work to be able to use the same module for replica.

PR #21, opened 2017-11-02
Automatically create enhanced monitoring role

s7anley:


s7anley:
Hi @antonbabenko, any chance this could be merged soon? 😉

antonbabenko:
Just came back from websummit and going to review it all during today/tomorrow. Sorry for the delay.

modules/db_instance/main.tf
```diff
@@ -1,6 +1,24 @@
...
 ##############
+
+locals {
+  create_enhanced_role = "${var.monitoring_role_autocreate != "" && var.monitoring_role_arn == ""}"
+}
+
+resource "aws_iam_role" "enhanced_monitoring" {
+  name               = "${var.monitoring_role_autocreate}"
+  assume_role_policy = "${file("${path.module}/policy/enhancedmonitoring.json")}"
+  count              = "${local.create_enhanced_role ? 1 : 0}"
```
antonbabenko: Put `count` as the first argument, add an empty line after it.

modules/db_instance/main.tf
```diff
@@ -1,6 +1,24 @@
...
+}
+
+resource "aws_iam_policy_attachment" "enhanced_monitoring" {
+  name       = "${var.monitoring_role_autocreate}-enhanced-monitoring-attachment"
+  roles      = ["${aws_iam_role.enhanced_monitoring.name}"]
+  policy_arn = "arn:aws:iam::aws:policy/service-role/AmazonRDSEnhancedMonitoringRole"
+  count      = "${local.create_enhanced_role ? 1 : 0}"
```
antonbabenko: Same here

README.md
```diff
@@ -42,8 +42,9 @@ module "db" {
   backup_window      = "03:00-06:00"
 
   # Enhanced Monitoring - see example for details on how to create the role
+  # by yourself, in case you don't want to create it automatically
   monitoring_interval = "30"
-  monitoring_role_arn = "arn:aws:iam::123456789012:role/rds-monitoring-role"
+  monitoring_role_autocreate = "rds-monitoring-role"
```
antonbabenko: `monitoring_role_autocreate` sounds like a boolean, not string.
Let's call this as `monitoring_role_name`, and use `create_monitoring_role` as the main switch - true/false.

modules/db_instance/main.tf
```diff
@@ -1,6 +1,24 @@
 ##############
 # DB instance
 ##############
+
+locals {
+  create_enhanced_role = "${var.monitoring_role_autocreate != "" && var.monitoring_role_arn == ""}"
```
antonbabenko: Avoid such rules (and `locals` in general unless it is possible to use specific module versions in Terraform Registry). Use `count = "${var.create_monitoring_role}"` as the main switch.

modules/db_instance/main.tf
```diff
@@ -1,6 +1,24 @@
...
+}
+
+resource "aws_iam_role" "enhanced_monitoring" {
+  name               = "${var.monitoring_role_autocreate}"
+  assume_role_policy = "${file("${path.module}/policy/enhancedmonitoring.json")}"
+  count              = "${local.create_enhanced_role ? 1 : 0}"
+}
+
+resource "aws_iam_policy_attachment" "enhanced_monitoring" {
+  name       = "${var.monitoring_role_autocreate}-enhanced-monitoring-attachment"
```
antonbabenko: `name = "${var.monitoring_role_name}"` - shorter is better usually. In this case, it is clear that it is an attachment of enhanced monitoring policy, so `-enhanced-monitoring-attachment` can be safely removed.

modules/db_instance/main.tf
```diff
@@ -31,7 +49,7 @@ resource "aws_db_instance" "this" {
   iops                = "${var.iops}"
   publicly_accessible = "${var.publicly_accessible}"
   monitoring_interval = "${var.monitoring_interval}"
-  monitoring_role_arn = "${var.monitoring_role_arn}"
+  monitoring_role_arn = "${coalesce(var.monitoring_role_arn, join("", aws_iam_role.enhanced_monitoring.*.arn))}"
```
antonbabenko: I like this hack with `join` 👍

variables.tf
```diff
@@ -114,6 +114,11 @@ variable "monitoring_role_arn" {
   default     = ""
 }
 
+variable "monitoring_role_autocreate" {
+  description = "Automatically create IAM role with a defined name that permits RDS to send enhanced monitoring metrics to CloudWatch Logs. Creation in skipped, when monitoring_role_arn is provided."
```
antonbabenko: `Creation is skipped`

README.md
s7anley: So I should additional variable `create_monitoring_role` for the switch and set some default name for `monitoring_role_name`?

modules/db_instance/main.tf
s7anley: I wanted to avoid the case of misconfiguration when a user provides ARN and also configure auto-creation, which could be skipped, but sure I will remove it.

README.md
antonbabenko: yes, and the defaults are:
```
create_monitoring_role = false
monitoring_role_name = ""
```

s7anley:
Thanks for CR, all your comments are addressed.

Issue #5, opened 2017-09-28
Add support for aws_db_option_group


antonbabenko:
Similar to this - https://github.com/douglasfraser-aviva/terraform-aws-rds/commit/a08d7039e151e17acacbf006ad17f3bfe04adb90

douglasfraser-aviva:
First attempt was non-functional, PR #24 is replacement.

antonbabenko:
[v1.17.0](https://github.com/terraform-aws-modules/terraform-aws-rds/releases/v1.17.0) has been released with this feature.

PR #24, opened 2017-11-16
Support database option group

douglasfraser-aviva:
Attempt to resolve add support for aws_db_option_group #5

### DIFF
--- a/modules/db_option_group/outputs.tf
+++ b/modules/db_option_group/outputs.tf
@@ -1,0 +1,10 @@
+# DB option group
+output "this_db_option_group_id" {
+  description = "The db option group id"
+  value       = "${aws_db_option_group.this.id}"
+}
+
+output "this_db_option_group_arn" {
+  description = "The ARN of the db option group"
+  value       = "${aws_db_option_group.this.arn}"
+}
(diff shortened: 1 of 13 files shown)


modules/db_option_group/outputs.tf
```diff
@@ -0,0 +1,10 @@
+# DB option group
+output "this_db_option_group_id" {
+  description = "The db option group id"
+  value       = "${aws_db_option_group.this.id}"
```
antonbabenko: Use this instead: `value       = "${element(split(",", join(",", aws_db_option_group.this.*.id)), 0)}"`

modules/db_option_group/outputs.tf
```diff
@@ -0,0 +1,10 @@
...
+}
+
+output "this_db_option_group_arn" {
+  description = "The ARN of the db option group"
+  value       = "${aws_db_option_group.this.arn}"
```
antonbabenko: And this - `value = "${element(split(",", join(",", aws_db_option_group.this.*.arn)), 0)}"`

modules/db_option_group/outputs.tf
zachawilson: @antonbabenko  This PR looks useful. Were these changes to output the only thing holding the PR back?

antonbabenko: Yes, I think so but also this PR should be rebased to use the latest version of the module.

antonbabenko:
[v1.17.0](https://github.com/terraform-aws-modules/terraform-aws-rds/releases/v1.17.0) has been released with this feature.

PR #25, opened 2017-11-16
Allow specification of timezone

douglasfraser-aviva:
timezone is supported by Microsoft SQL Server, allow it as optional parameter.


antonbabenko:
Thanks!

Could you also update examples? Add mssql example with `timezone` specified.

Also, make sure that existing examples works. They should work, but worth checking.

antonbabenko:
ping @douglasfraser-aviva

antonbabenko:
Resolved in #51

PR #27, opened 2017-11-17
Var with default value in order to prevent the error

agomezvidalee:
Error applying plan:

1 error(s) occurred:

* module.database.module.db_instance.aws_db_instance.this (destroy): 1 error(s) occurred:

* aws_db_instance.this: InvalidParameterValue: The parameter FinalDBSnapshotIdentifier is not a valid identifier. Identifiers must begin with a letter; must contain only ASCII letters, digits, and hyphens; and must not end with a hyphen or contain two consecutive hyphens.
	status code: 400, request id: xxxxx

Terraform does not automatically rollback in the face of errors.
Instead, your Terraform state file has been partially updated with
any resources that successfully completed. Please address the error
above and apply again to incrementally change your infrastructure.


agomezvidalee:
duplicated

PR #30, opened 2017-12-05
Add create_before_destroy to parameter group

dpiddockcmp:
aws_db_parameter_group (or AWS?) does not support removing previously
defined parameters.

I thought I'd give "taint" a spin and force the parameter group to be
recreated.  terraform tried to delete the parameter group but failed as
it's still in use.

This change allows terraform to migrate the RDS instance to the new
parameter group before destroying the old.


antonbabenko:
Thanks, [v1.6.0](https://github.com/terraform-aws-modules/terraform-aws-rds/releases/tag/v1.6.0) is released.

PR #33, opened 2017-12-22
You are asking for kms_key_id, but what you really are asking for is kms_key_arn

pollosp:
Hi, 
It looks like you are asking for kms_key_id, but what you really are asking for is kms_key_arn


antonbabenko:
According to [Terraform documentation](https://www.terraform.io/docs/providers/aws/r/db_instance.html#kms_key_id) it should be `kms_key_id`, so closing this.

PR #34, opened 2018-01-15
Fix enhanced monitoring policy attachment

nazartm:


antonbabenko:
Thanks, worth mentioning that this is exactly how it enhanced monitoring is shown in the example code.

antonbabenko:
v1.7.0 is released.

PR #35, opened 2018-01-23
Add availability zone to instance ressource module

n3ph:
We use db_instance module directly and need to define the availability zone per instance.


antonbabenko:
[v1.8.0](https://github.com/terraform-aws-modules/terraform-aws-rds/releases/v1.8.0) is released

PR #36, opened 2018-01-26
Support conditional creation for the database too.

jimrazmus:
I want to be able to conditionally create databases. For example, I may only want read replica databases in the staging and production environments. Lower tiered environments would not have a read replica. The count variable provides that flexibility while maintaining the DRY principle. In the absence of this variable, I have to maintain multiple, environment specific, terraform files.

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -4,6 +4,7 @@
 module "db_subnet_group" {
   source = "./modules/db_subnet_group"
 
+  count       = "${var.create_db_subnet_group}"
   identifier  = "${var.identifier}"
   name_prefix = "${var.identifier}-"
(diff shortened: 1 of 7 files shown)


antonbabenko:
Good work, but adjust outputs when `count` is set to 0 in one of the recent version of AWS provider. Something like - `element(concat(..., list("")), 0)`

jimrazmus:
@antonbabenko I believe this PR is now ready for consideration. Thank you!

antonbabenko:
Thanks! Always run at least one of examples to reveal the broken parts, I have updated outputs to fix it.

New release is [v1.10.0](https://github.com/terraform-aws-modules/terraform-aws-rds/releases/v1.10.0).

PR #38, opened 2018-02-08
Enable db_subnet_group_name variable

mhdali:
If the variable not specify then use the db_subnet_group module

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -1,3 +1,7 @@
+locals {
+  db_subnet_group_name = "${coalesce(var.db_subnet_group_name, module.db_subnet_group.this_db_subnet_group_id)}"
+}
+
 ##################
 # DB subnet group
 ##################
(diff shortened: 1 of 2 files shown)


main.tf
```diff
@@ -56,7 +60,7 @@ module "db_instance" {
   snapshot_identifier = "${var.snapshot_identifier}"
 
   vpc_security_group_ids = ["${var.vpc_security_group_ids}"]
-  db_subnet_group_name   = "${module.db_subnet_group.this_db_subnet_group_id}"
+  db_subnet_group_name   = "${local.db_subnet_group_name}"
```
antonbabenko: When `var.db_subnet_group_name` is not empty it will still respect `var.create_db_subnet_group` and probably create unused db_subnet_group.

Can you change argument `count` in `db_subnet_group` ([line 11](https://github.com/terraform-aws-modules/terraform-aws-rds/pull/38/files#diff-7a370d8342e7203b805911c92454f0f4R11))  to respect `var.create_db_parameter_group` only if `var.db_subnet_group_name` is not set?

mhdali: 👍

mhdali: Initially I put the condition on `count` value, but it felt a bit complicated to I moved it to locals.

Let me know if you want to move it back to `count`.

antonbabenko:
Looks flexible enough. Good work.

antonbabenko:
[v1.11.0](https://github.com/terraform-aws-modules/terraform-aws-rds/releases/v1.11.0) has been released.

PR #47, opened 2018-03-09
Enable db_parameter_group_name variable

jordiclariana:
If the variable not specify then use the db_parameter_group module
This mimics the work done in PR #38 about `db_subnet_group_name`


mazzy89:
Could we have this reviewed? I have the use case where the parameter group already exists and I need to pass in as a variable rather than create a new one. It would be very useful. For me it looks fine the job done by @jordiclariana

antonbabenko:
[v1.14.0](https://github.com/terraform-aws-modules/terraform-aws-rds/releases/v1.14.0) has been just released. Thank you for the PR and a nudge :)

PR #50, opened 2018-03-20
Add apply_method to parameter group

tornadical:
Adding [apply_method](https://www.terraform.io/docs/providers/aws/r/db_parameter_group.html#apply_method) variable to parameter group to allow for parameters that can't be applied immediately.


antonbabenko:
`apply_method` should be part of `parameter` block and not part of `aws_db_parameter_group` resource. Read [this](https://www.terraform.io/docs/providers/aws/r/db_parameter_group.html). Closing.

PR #77, opened 2018-08-28
Mark parameter and option group vars as required

tlvince:
`family` (DB parameter group) and `major_engine_version` (DB option group) are both required (in their corresponding module). Trying to apply without these set throws an error.


Vlaaaaaaad:
Is there a reason this hasn't been merged yet? The change is not that large and the module does indeed not work if those variables are not given.

dpiddockcmp:
`family` is not required if `parameter_group_name` is set or `create_db_parameter_group` is `false`

`major_engine_version` is not required if `option_group_name` is set or `create_db_option_group` is `false`

This "simple change" is a backward-incompatible breaking change.

Vlaaaaaaad:
Oooooh. Makes perfect sense!

MrSaints:
> family is not required if parameter_group_name is set or create_db_parameter_group is false

FWIW, that should probably be added in the `README.md`.

PR #81, opened 2018-09-24
final_snapshot_identifier has bad default #57

### LINKED ISSUE #57: final_snapshot_identifier has bad default
./modules/db_instance/variables.tf final_snapshot_identifier is defined as
```
variable "final_snapshot_identifier" {
  description = "The name of your final DB snapshot when this DB instance is deleted."
  default     = false
}
```
According to terraform docs `false` will convert to the string `0`
https://www.terraform.io/docs/configuration/variables.html

Thus we are trying to to have a final snapshot with an identifier of 0. This however produces 
```
* aws_db_instance.this: InvalidParameterValue: The parameter FinalDBSnapshotIdentifier is not a valid identifier. Identifiers must begin with a letter; must contain only ASCII letters, digits, and hyphens; and must not end with a hyphen or contain two consecutive hyphens.
```

ozbillwang:
Fix #57 

a quick fix that `false` can't be used as aws rds final snapshot identifier.

### DIFF
--- a/variables.tf
+++ b/variables.tf
@@ -51,7 +51,7 @@ variable "engine_version" {
 
 variable "final_snapshot_identifier" {
   description = "The name of your final DB snapshot when this DB instance is deleted."
-  default     = false
+  default     = ""
 }
 
 variable "instance_class" {


ozbillwang:
If #82 can be approved and merged, this PR (#81) is not required any more.

bryantbiggs:
this was changed to `null` on #126

PR #82, opened 2018-09-24
#82 - generate final snapshot identifier name and avoid DBSnapshotAlreadyExists issue

ozbillwang:
For the same database (same rds name), if you destroy the rds instance second time with same code (same final_snapshot_identifier), you will get below error:

>aws_db_instance.aws_db_instance: error deleting Database Instance "test-dev": DBSnapshotAlreadyExists: Cannot create the snapshot because a snapshot with the identifier test-dev-final-snapshot already exists.

To avoid this issue and simplify the codes, I set `final_snapshot_identifier` from rds instance name with a random string. 

let me know if you like this idea.

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,9 +61,6 @@ module "db" {
   # DB option group
   major_engine_version = "5.7"
 
-  # Snapshot name upon DB deletion
-  final_snapshot_identifier = "demodb"
-
   parameters = [
     {
       name = "character_set_client"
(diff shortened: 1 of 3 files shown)


dud225:
The md5 computation is not stable so TF computes a new plan at every run.
I suggest to use a random_id resource.

bryantbiggs:
this was changed to `null` on #126

Issue #90, opened 2018-11-16
Defaults for 'skip_final_snapshot' and 'backup_retention_period' should align with AWS API


lvdh:
README.md has the following snapshot config in its example:

```
  # Snapshot name upon DB deletion
  final_snapshot_identifier = "demodb"
```

However, setting the `final_snapshot_identifier` while `skip_final_snapshot` has a default value of `true` means that this final snapshot will never take place.

This behavior is confusing. Additionally, it does not align with the [AWS API default of `skip_final_snapshot=false`](https://docs.aws.amazon.com/AmazonRDS/latest/APIReference/API_DeleteDBInstance.html). Neither does the `backup_retention_period` (now 1, AWS default = 7)

lvdh:
Submitted https://github.com/terraform-aws-modules/terraform-aws-rds/pull/91

PR #91, opened 2018-11-16
Fix RDS snapshot defaults

lvdh:
Cf. https://github.com/terraform-aws-modules/terraform-aws-rds/issues/90

### DIFF
--- a/variables.tf
+++ b/variables.tf
@@ -166,7 +166,7 @@ variable "copy_tags_to_snapshot" {
 
 variable "backup_retention_period" {
   description = "The days to retain backups for"
-  default     = 1
+  default     = 7
 }
 
 variable "backup_window" {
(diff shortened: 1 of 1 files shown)


variables.tf
```diff
@@ -156,7 +156,7 @@ variable "maintenance_window" {
 
 variable "skip_final_snapshot" {
   description = "Determines whether a final DB snapshot is created before the DB instance is deleted. If true is specified, no DBSnapshot is created. If false is specified, a DB snapshot is created before the DB instance is deleted, using the value from final_snapshot_identifier"
-  default     = true
+  default     = false
```
vutny: :+1: for this! It would be in line with the Terraform resource defaults:
https://www.terraform.io/docs/providers/aws/r/db_instance.html#skip_final_snapshot

Also, it makes sense that the `final_snapshot_identifier` input is declared as required for the module in the registry:
https://registry.terraform.io/modules/terraform-aws-modules/rds/aws/2.14.0?tab=inputs#required-inputs

PR #98, opened 2019-03-04
mark the database password as sensitive

mattyjones:
When doing a plan or when an error log is produced the database password is exposed. This should not be the case and I believe marking this as sensative will solve this according the current [docs](https://www.terraform.io/docs/configuration/outputs.html#sensitive-outputs).


bryantbiggs:
@mattyjones would you mind rebasing to fix merge conflicts when you get a chance?

antonbabenko:
@mattyjones [v2.8.0](https://github.com/terraform-aws-modules/terraform-aws-rds/releases/v2.8.0) has been just released with this minor fix.

Issue #103, opened 2019-03-12
Error when creating a read replica


carlosfrancia:
This modules returns an error while creating a Read Replica in both cases, same region and Cross-Region. 

In order to create a replica the variable replicate_source_db it must be populated. I.e:

```
replicate_source_db = "${module.db.this_db_instance_arn}"
```

The aws_rds_resource doesn't require of username and password variables when being an instance of a read replica (https://www.terraform.io/docs/providers/aws/r/db_instance.html):

_username - (Required unless a snapshot_identifier or replicate_source_db is provided) Username for the master DB user._

Trying to use this module without Username or password fails on terraform init:

```
Error: module "db-crossregion-replica": missing required argument "username"

Error: module "db-crossregion-replica": missing required argument "password"
```

Trying to add an empty password and get the following error on terraform plan:

```
Error: Error running plan: 1 error(s) occurred:

* module.db-replica.module.db_instance.local.this_db_instance_password: local.this_db_instance_password: Resource 'aws_db_instance.this' does not have attribute 'password' for variable 'aws_db_instance.this.*.password'
```
So password needs to be populated with some value and an error is returned on terraform apply:
 
```
Error: Error applying plan:

1 error(s) occurred:

* module.db-replica.module.db_instance.aws_db_instance.this: 1 error(s) occurred:

* aws_db_instance.this: error modifying DB Instance (demodb-replica): InvalidParameterCombination: Cannot change master user password on an RDS Postgres Read Replica because it uses physical replication and therefore cannot differ from its parent.
	status code: 400, request id: 794f2fc6-9f70-442a-acc6-5724b81b2d0c
```

**Expected result**

A Read Replica is created without errors

**Actual result**

Replica is created but it returns an error.

adv4000:
Same error here, Replica created but got error  `Cannot change master user password on an RDS Postgres Read Replica because it uses physical replication and therefore cannot differ from its parent` 

After additional run of **terraform apply**  it finish without error.

antonbabenko:
I believe this is what @dpiddockcmp has fixed in #108 and it has been released as [v1.27.0](https://github.com/terraform-aws-modules/terraform-aws-rds/releases/v1.27.0) an hour ago.

Please let us know if there is something still missing (or not working as expected).

PR #104, opened 2019-03-13
Adding support for Read Replicas creation

carlosfrancia:
Changes to fix issue https://github.com/terraform-aws-modules/terraform-aws-rds/issues/103

Added a new module to create Read Replicas in the same region or in a different one based on a new variable Is_replica (default value is false). 

The problem in the current version is that the terraform resource aws_db_instance is always created with username and password variables. However, for read replicas username and password should not be used (https://www.terraform.io/docs/providers/aws/r/db_instance.html):

> password - (Required unless a snapshot_identifier or replicate_source_db is provided)


antonbabenko:
I believe this is what @dpiddockcmp has fixed in #108 and it has been released as [v1.27.0](https://github.com/terraform-aws-modules/terraform-aws-rds/releases/v1.27.0) an hour ago.

Closing this pull-request because the same functionality is already implemented.

Please let us know if there is something still missing (or not working as expected).

PR #108, opened 2019-03-22
Remove unused submodule output

dpiddockcmp:
This breaks plan when trying to create a read replica without setting
the master password:

`* module.database.module.db_instance.local.this_db_instance_password: local.this_db_instance_password: Resource 'aws_db_instance.this' does not have attribute 'password' for variable 'aws_db_instance.this.*.password'`

Less invasive solution for #103 . Although I'm working with MariaDB not Postgres.

Need to set password and username to empty strings. Maybe will be better under TF 0.12 with a null value?

### DIFF
--- a/examples/replica-postgres/main.tf
+++ b/examples/replica-postgres/main.tf
@@ -1,0 +1,99 @@
+provider "aws" {
+  region = "eu-west-1"
+}
+
+####################################
+# Variables common to both instnaces
+####################################
+locals {
+  engine            = "postgres"
+  engine_version    = "9.6.3"
+  instance_class    = "db.t2.large"
+  allocated_storage = 5
+  port              = "5432"
+}
+
+##############################################################
+# Data sources to get VPC, subnets and security group details
+##############################################################
+data "aws_vpc" "default" {
+  default = true
+}
+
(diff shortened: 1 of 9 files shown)


antonbabenko:
I think it is a good idea to remove the password from the output, but I am not sure why it makes problems now and not before.

Could you show the code you are trying to get working and add it as a new example to `examples` directory? I think it is a pretty good idea to show how read-replicas can be created.

dpiddockcmp:
Hi @antonbabenko. I've added simple examples for mysql and postgresql. These fail against master but work with the original changes here.

examples/replica-postgres/main.tf
```diff
@@ -0,0 +1,99 @@
...
+
+############
+# Replica DB
+############
+module "replica" {
+  source = "../../"
+
+  identifier = "demodb-replica"
+
+  # Source database. For cross-region use this_db_instance_arn
```
antonbabenko: 👍

antonbabenko:
Thanks, Daniel!

[v1.27.0](https://github.com/terraform-aws-modules/terraform-aws-rds/releases/v1.27.0) has been released.

PR #117, opened 2019-04-29
Add lifecycle block ignore_changes attributes for db_instance

robinbowes:
We need to set a lifecycle block on the db_instance to ignore changes to the snapshot_identifier.

This PR allows the "ignore_changes" list of attributes to be passed in at the top level, and is passed down to the db_instance module.

The default value is an empty list.


robinbowes:
Damn, this runs into https://github.com/hashicorp/terraform/issues/3116

Closing.

antonbabenko:
Hi @robinbowes !

Yes, unfortunately, it is still an open issue in Terraform.

Issue #42, opened 2018-02-16
db_subnet_group support for name?


allaire:
Hi, I see that the `db_subnet_group` module only support `name_prefix` and not `name`. Can you shed some lights on that decision? I could probably create manually my db subnet group and pass its name to `db_subnet_group_name`, but is it recommended to have a db subnet group per rds instance, hence why name_prefix is only supported?

Thanks!

antonbabenko:
`name_prefix` helps to create unique resources, which is important when reusing this module multiple times within the same scope (eg, single region in single AWS account where resources should be unique). `name_prefix` is preferred over `name` for the majority of resources, not just for RDS.

danvaida:
@antonbabenko what about people that want handle the "uniqueness" themselves by using "name"?

antonbabenko:
@danvaida Hi! :)

This should be possible to implement easier once Terraform 0.12 is ready. Currently, this module does not have this feature.

PR #121, opened 2019-05-21
Use name or name_prefix consistently

cytopia:
## Use `name` or `name_prefix` consistently

This PR ensures that the user can consistently choose to either use `name` or `name_prefix` for all sub-module resources:

* Option group (via this PR)
* Subnet group (via this PR)
* Parameter group (already implemented)

It should be a non-breaking change, as the default behaviour remains (using `name_prefix`).

It will additionally fix this issue:
* Fixes: #42


bryantbiggs:
@cytopia would you mind rebasing to fix merge conflicts when you get a chance?

ronaldour:
Do you have plans of merging this PR? we are facing issues that this PR will probably fix

PR #123, opened 2019-05-28
Terraform 0.12 syntax changes

wardviaene:
Fixes #111

PR #126, opened 2019-06-11
 Upgraded module to support Terraform 0.12

antonbabenko:
Based on #123

Thanks to everyone involved in this process!


antonbabenko:
If you are using Terraform 0.12, you can use [v2.0.0](https://github.com/terraform-aws-modules/terraform-aws-rds/releases/v2.0.0) of this module which has been just released.

PR #147, opened 2019-07-16
Add support for performance insights retention period

walkafwalka:


antonbabenko:
Thanks @walkafwalka !

[v2.4.0](https://github.com/terraform-aws-modules/terraform-aws-rds/releases/v2.4.0) has been just released.

mprenditore:
Hello,

I think there is a bug in this PR, if the `performance_insight_enabled` is set to `false` as per default I'm getting this error while executing the job:

```Error: Error modifying DB Instance <<db>>: InvalidParameterCombination: Cannot set PerformanceInsightsRetention without EnablePerformanceInsights
	status code: 400, request id: <<request_id>>```

As a simple fix, due the not programmability of TF, we could duplicate the `aws_db_instance` resource  and in one add the definition of the variables `performance_insight_enabled` and `performance_insight_retention_period` and in the other we don't add it.
To discriminate about which one to apply we can define `count` var to:

- `count = var.create && local.is_mssql && car.performance_insight_enabled ? 1 : 0`
- `count = var.create && local.is_mssql && car.performance_insight_enabled == false ? 1 : 0`

I don't know if there is a better way to do it in TF, I'm quite new to it.

Thanks,

Stefano

PR #168, opened 2019-10-01
Support for EC2-Classic RDS using security_group_names

rmkbow:
# Description

Adding security group name parameter to support EC2-Classic RDS instances


antonbabenko:
Hi Ronald,

While I understand the reason for this change, I don't think we need to do an effort to support EC2-Classic features in this module. EC2-Classic has been already marked as not the recommended way of using AWS by AWS for a few years (yes, there are still companies using EC2-Classic).

If you need to have this feature, please use your own fork.

Closing this issue.

PR #169, opened 2019-10-08
Add optional security group.

evanstachowiak:
# Description

I added a feature that adds an optional security group. Every rds needs a security group to function, so I thought this was a nice option to add. It is not enabled by default, to foster backwards comparability, and can be enabled with `var.sg_create` set to `true`.

Open to feedback and suggestions here.

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -1,3 +1,7 @@
+data "aws_subnet" "selected" {
+  id = var.subnet_ids[0]
+}
+
 locals {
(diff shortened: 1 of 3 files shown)


bryantbiggs:
I'm not sure this makes a lot of sense. If you look at some of the other official modules ([asg](https://github.com/terraform-aws-modules/terraform-aws-autoscaling), [ec2-instance](https://github.com/terraform-aws-modules/terraform-aws-ec2-instance)), they also follow the approach of "bring your own security group". There are a number of issues with baking in the security group, not withstanding the embedding of an external module within the module.

antonbabenko:
I agree with @bryantbiggs . Closing this PR.

Issue #107, opened 2019-03-18
aws_db_option_group.this: value of 'count' cannot be computed


kanewinter:
When option_group_name is set it causes this error:
 module.rds.module.db_option_group.aws_db_option_group.this: aws_db_option_group.this: value of 'count' cannot be computed
This makes it impossible to select a pre-existing option group.

kanewinter:
`module "rds" {
  source = "terraform-aws-modules/rds/aws"
  identifier = "rds-${var.environment}"

 
  engine            = "${var.rds_engine}"
  engine_version    = "${var.rds_engine_version}"
  instance_class    = "${var.rds_instance_class}"
  allocated_storage = "${var.rds_allocated_storage}"
  storage_encrypted = "${var.rds_storage_encrypted}"
  multi_az          = "${var.rds_multi_az}"

  # kms_key_id        = "arm:aws:kms:<region>:<accound id>:key/<kms key id>"
  name     = "rds_${var.environment}"
  username = "rds_${var.environment}"
  password = "${var.rds_password}"
  port     = "${var.rds_port}"
  vpc_security_group_ids = ["${data.terraform_remote_state.state.default_sg_id}"]
  maintenance_window = "Mon:00:00-Mon:03:00"
  backup_window      = "03:00-06:00"

  # disable backups to create DB faster
  backup_retention_period = "${var.rds_retention}"

  tags = "${merge(map(
    "Name", "rds_${var.environment}"),
    var.global_tags
  )}"

  # DB subnet group
  subnet_ids = ["${data.terraform_remote_state.global_state.dev_subnet_ids}"]

  # DB parameter group
  family = "mysql5.7"

  # DB option group
  major_engine_version = "5.7"

  # Snapshot name upon DB deletion
  final_snapshot_identifier = "rds_dev_db"

  parameters = [
    {
      name = "log_bin_trust_function_creators"
      value = true
    }
  ]

  create_db_option_group = false
  option_group_name = "${aws_db_option_group.default.id}"
}`

antonbabenko:
Fixed in [v2.11.0](https://github.com/terraform-aws-modules/terraform-aws-rds/releases/v2.11.0).

PR #170, opened 2019-10-14
Fix `enable_create_db_option_group` to use override first & computed second

mattclegg:
This fixes issue; https://github.com/terraform-aws-modules/terraform-aws-rds/issues/107

By explicitly stating a value for `create_db_option_group` then you can use a computed value for `option_group_name`.


dre1992:
This change actually solved the problem for us

antonbabenko:
Thanks, Matt!

[v2.11.0](https://github.com/terraform-aws-modules/terraform-aws-rds/releases/v2.11.0) has been released with this change.

PR #176, opened 2019-12-06
Downcase params when required

shoekstra:
Some fields require the value to be lowercase, however the variables used are used in other places so it makes sense to only downcase where required.

### DIFF
--- a/modules/db_instance/main.tf
+++ b/modules/db_instance/main.tf
@@ -39,7 +39,7 @@ resource "aws_iam_role_policy_attachment" "enhanced_monitoring" {
 resource "aws_db_instance" "this" {
   count = var.create && false == local.is_mssql ? 1 : 0
 
-  identifier = var.identifier
+  identifier = lower(var.identifier)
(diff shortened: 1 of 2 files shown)


antonbabenko:
Hi @shoekstra !

While this sounds like a good idea for some cases, I don't think that this module should do any manipulations with the data inputs unless it does not work without it. I really prefer to keep this module plain and avoid any kind of behind-the-scene transformations.

shoekstra:
Hello @antonbabenko !

Thanks for the feedback and apologies for the delayed response.

> I don't think that this module should do any manipulations with the data inputs unless it does not work without it. 

In this instance I am doing a manipulation to make the plan work; the input does not work unless it is down cased, so I'm only amending the modules in the in required places in order to make it work.

It felt wasteful to declare two variables with almost the same content (one being capitalised) in order to comply with the API.  Since the API requires the string to be lowercase, it makes sense (IMHO) to downcase the value when required.

PR #177, opened 2019-12-15
feat: add Windows authentication

bryantbiggs:
# Description

- Add Windows authentication per https://www.terraform.io/docs/providers/aws/r/db_instance.html\#domain which is now supported by all RDS engines

closes #193

PR #175, opened 2019-12-03
add support for ca_cert_identifier option

carlo-blohm:
This PR adds the option to set the ca_cert_identifier
As the current CA is expiring 5th March 2020 i need to update my clusters to use the new CA before that. The default is currently set to the old one to not break any existing installations.

This will fix #171

https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/UsingWithRDS.SSL-certificate-rotation.html

PR #145, opened 2019-07-10
Added delete timeout for db_option_group [#144]

Constantin07:
This PR addresses the issue in #144 

It adds possibility to specify delete timeout more than 15 mins for db_option_group.
AWS takes sometime more than default - thus failing the terraform plan and requiring to re-run it.

It doesn't (should not change) the existing behaviour (non-breaking change).

PR #115, opened 2019-04-24
add output for aws_db_instance

SolidSly:
Cf. #114

PR #178, opened 2019-12-16
Feature / Combined Updates

bryantbiggs:
# Description

- Add MSSQL active directory auth per https://www.terraform.io/docs/providers/aws/r/db_instance.html\#domain
- Bump pre-commit versions and run pre-commit to generate documentation changes
- Mark database password output as sensitive to avoid exposing in stdout output
- Correct usage of `name` and `name_prefix` on option group, subnet group, and parameter group. Previously all defaulted to use a prefix but there are now 3 scenarios supported (as was the original intent I believe). a user can provide a name to an existing group to use (create flag = false), a user can provide a name to create a new group (create flag = true), a user can provide a name prefix to create a new group (create flag = true && use prefix = true). 
- Add support for certificate authority setting and its associated output
- Add delete timeout configuration for option group, currently set to resource default
- Subnet group, option group, and parameter group names and the database instance identifier have been made lowercase as this is either required and/or when created manual they will be converted to lowercase (tested through console)
- The use of name, name prefix, and identifier have been updated to be consistent. when a group name is not provided, the db instance identifier will be used, and likewise for name prefix. the `Name` tag was updated to be consistent with what is selected from the above coalesce results to be consistent

This PR is rather quite large but that is due to the culmination of changes that have been building up in open PRs that have gone stale, did not function as intended, etc. I have tested and validated these changes and they appear to be working as intended - currently this setup is used in production on a MSSQL instance without issue. Hopefully we can get this resolved quickly, let me know if I can be of any help!

Note - at minimum this supersedes and closes the following PRs if merged: #177, #98, #121, #115, #175, #145 - there may be more but that can be determined after resolution

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,10 +1,10 @@
 repos:
 - repo: git://github.com/antonbabenko/pre-commit-terraform
-  rev: v1.17.0
+  rev: v1.21.0
   hooks:
     - id: terraform_fmt
     - id: terraform_docs
 - repo: git://github.com/pre-commit/pre-commit-hooks
-  rev: v2.2.3
+  rev: v2.4.0
   hooks:
     - id: check-merge-conflict
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,9 +3,75 @@
 
 
 
+<a name="v2.8.0"></a>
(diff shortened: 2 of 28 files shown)


bryantbiggs:
@antonbabenko let me know if there is anything I can do to help speed up the review process and get to a resolution, thank you!

PR #174, opened 2019-11-25
Add support for ca_cert_identifier option

smelchior:
# Description

This PR adds the option to set the ca_cert_identifier
As the current CA is expiring 5th March 2020 i need to update my clusters to use the new CA before that. The default is currently set to the old one to not break any existing installations. 

This will fix #171

https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/UsingWithRDS.SSL-certificate-rotation.html

PR #179, opened 2019-12-24
Missing `ca_cert_identifier` for sqlserver engine aws rds instance

buevichp:
The recent PR (https://github.com/terraform-aws-modules/terraform-aws-rds/pull/174) included changes for the `ca_cert_identifier` variable, but it does not have it on the `aws_db_instance.this_mssql` resources (sqlserver engine types). This may be important for some AWS users. Also, we've briefly tested that the change can be applied without destroying the resource, but requires `apply-immediately` to be set .

# Description

Please explain the changes you made here and link to any relevant issues.

### DIFF
--- a/modules/db_instance/main.tf
+++ b/modules/db_instance/main.tf
@@ -161,6 +161,8 @@ resource "aws_db_instance" "this_mssql" {
   backup_window           = var.backup_window
 
   timezone = var.timezone
+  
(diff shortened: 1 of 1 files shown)


buevichp:
I think this should be closed in favor to https://github.com/terraform-aws-modules/terraform-aws-rds/pull/178

PR #189, opened 2020-01-11
Fix / CA Cert Instance Default

bryantbiggs:
# Description

- Update `db_instance` sub-module to use same default ca_cert_identifier as the root level module default

PR #196, opened 2020-02-20
Set ca_cert_identifier default for internal module

robinbowes:
# Description

Same as #189  but for terraform011 branch.

Sets default value for ca_cert_identifer for internal module.

Also, some terraform-docs re-formatting


antonbabenko:
Thanks @robinbowes for the great support!

[v1.37.0](https://github.com/terraform-aws-modules/terraform-aws-rds/releases/v1.37.0) has been just released.

PR #209, opened 2020-03-25
feat: conditional use of nameprefix for subnet and option groups

marcportabellaclotet-mt:
## Description
In the current modules, name prefix is always created for option groups and subnet groups. However for parameter groups there is a conditional variable (use_name_prefix) to decide wether to use nameprefix or not
This PR adds the option to choose if name prefix is used when creating subnet groups and option groups.

## Motivation and Context
The motivation of this change is to be able to import current infrastructure to terraform without recreating resources that are not using nameprefix

## Breaking Changes
None


s4kharitonov:
Will this pull request be accepted? I'm trying to import existing subnet and option group that not have name_prefix and getting "forces replacement"

factotum-fox:
Could you please pull this? This is an interesting new feature

sergirosell:
this is also requested on a new issue: https://github.com/terraform-aws-modules/terraform-aws-rds/issues/230

do you have any prevision for when this request will be accepted and merged? @antonbabenko

antonbabenko:
@sergirosell I believe that my comment from above https://github.com/terraform-aws-modules/terraform-aws-rds/pull/209#pullrequestreview-410773960 is still relevant and I will be glad to merge it once it is fixed by somebody.

PR #210, opened 2020-03-30
Updated README

kompaniets-viacheslav:
Family DB parameter group is required

## Description
During RDS instance creation parameter Family (Parameter Group) was not provided, because in readme this parameter marked as not required.
Error that appears: 
```
Error: Error creating DB Parameter Group: InvalidParameterValue: The parameter ParameterGroupFamily must be provided and must not be empty.
	status code: 400, request id: 4a49edc1-981d-4d08-YYYY-XXXXXXXXXXXX
```
Changed readme: Family option now required.

## Motivation and Context
I faced with an issue during RDS creation process but mentioned that Readme says that this  parameter is not reqiured.

## Breaking Changes
No

## How Has This Been Tested?
By applying


antonbabenko:
The parameter is marked as not required because user may want to not create the parameter group thus it should not be specified.

This PR can be closed.

PR #216, opened 2020-04-12
feat: Add support for character_set_name with MSSQL Server

alo-is:
## Description
Added support for character_set_name (also known as collation) in Microsoft SQL Server on RDS. This feature is optional, but supported in MSSQL on RDS.

## Motivation and Context
Without this attribute, newly created MSSQL databases on RDS will be set to SQL_Latin1_General_CP1_CI_AS, which is the default value, but is incorrect in some cases when specific charsets are used, or with some software (we've got the issue with Sage Accounting solution).
The full AWS documentation regarding this attribute is available here : https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/Appendix.SQLServer.CommonDBATasks.Collation.html

## Breaking Changes
No impact on current version, if is attribute is left empty, then default value will apply as before (set by AWS, not this module).

## How Has This Been Tested?
This change has been tested by deploying MSSQL 2017 Servers on RDS (express + web version). With & without the attribute, the changes are working as expected, changing the database default collation to the wanted one.


bryantbiggs:
closed with #281

PR #226, opened 2020-05-13
fix: Update minimum provider version to account for ca_cert_identifier

### LINKED ISSUE #224: Update provider version to reflect requirements
The provider requirement for Terraform 0.12.x appears to be lower than what is necessary to use this module. Per #184, which I experienced today, provider version should be `~> 2.37`

https://github.com/terraform-aws-modules/terraform-aws-rds/blob/9aff82bec1bcfe1c7a9f5479bfda7926509bd967/versions.tf#L4-L6

rafmagns-skepa-dreag:
## Description
Updated the minimum version of the AWS provider so that `ca_cert_identifier` could be set.

## Motivation and Context
The current minimum version appears to work for everything else but was not updated when adding support for `ca_cert_identifier`. This causes errors when users meet the requirements laid out in the file but not
Fixes #224 

## Breaking Changes
This should not break backwards compatibility.

### DIFF
--- a/versions.tf
+++ b/versions.tf
@@ -2,6 +2,6 @@ terraform {
   required_version = "~> 0.12.6"
 
   required_providers {
-    aws = "~> 2.23"
+    aws = "~> 2.37"
   }
 }


antonbabenko:
Thanks @rafmagns-skepa-dreag !

[v2.15.0](https://github.com/terraform-aws-modules/terraform-aws-rds/releases/v2.15.0) has been just released.

PR #251, opened 2020-08-24
fix: Do not create option_group if create_db_option_group = false

odavid:
## Description

Currently `db_option_group` is being created despite the fact `create_db_option_group = false` 
The current condition does not respect the  `create_db_option_group` if `option_group_name` is not equal and `var.engine` equals `postgres`

## Motivation and Context

I believe passing create_option_group as false must not create an option_group

## Breaking Changes

No breaking changes


bronhy:
I believe this commit does not solve the issue. 

if my input is 
```
var.create_db_option_group = true
var.option_group = "my-test-option-group"
```

the code below will produce false
```
enable_create_db_option_group = var.create_db_option_group && var.option_group_name == ""
```

There is a parallel pull request which would fix it  https://github.com/terraform-aws-modules/terraform-aws-rds/pull/222/commits/bdc2de20ced43490d02f61be4c704eae282785fb

PR #252, opened 2020-09-09
Identifier should not be required

holyjak:
...according to its docs


bryantbiggs:
hi @holyjak - correct, it is not a required value per the db instance resource itself, in isolation - however, this value was used (and still is) in multiple places and therefore this module does require a value to be provided for `identifier`

PR #254, opened 2020-09-24
test: Integration test for modules postgres and myslq complete

silva-jailson:
## Description
Add automated test with terratest to modules complete-mysql and complete-postgres.

Plus - Add detect-aws-credentials as hook on pre-commit 

## Motivation and Context

The goal is to make testing and validate faster before releases of versions.

To make that test works some changes was made on examples folder:

 - Moved some parameters from main.tf to variables.tf

 - To connect on database, I need to set RDS Yes for public accessibility, so to keep the test safe, I changed the default SG, now I create a new SG and set the public IP this way only those who run the test can access it.

## Breaking Changes

No breaking changes

## If you wants run all the tests

```bash
cd test
go test -v -timeout 45m
```

## If you wants run a specific test

To run a specific test called `TestTerraformAwsCompleteMysql`:

```bash
cd test
go test -v -timeout 45m -run TestTerraformAwsCompleteMysql
```


bryantbiggs:
thank you for the PR @silva-jailson - however, we will not be introducing terratest into the modules at this time

PR #255, opened 2020-09-30
fix: Make this_db_instance_* output usable in providers

modax:
It appears that zero-length splat resources will break early
terraform dependency resolution (e.g. which is used for instantiating
terraform providers). For example, `this_db_instance_*` outputs (esp.
this_db_instance_endpoint) are useful in mysql provider. So the fix
makes something like the following work:

```
module "db" {
  source  = "terraform-aws-modules/rds/aws"

  ...
  username = var.db_user
  password = var.db_password
  port     = var.db_port
  ...
}

provider "mysql" {
  endpoint = module.db.this_db_instance_endpoint
  username = var.db_user
  password = var.db_password
}

resource "mysql_database" "db" {
  name = "testdb"
}
```

This commit adds short circuiting conditional (between `this_mssql` and
`this` aws_db_instance) for this_db_instance_* outputs hence avoiding
evaluation of zero-length splat aws_db_instance resources.


bryantbiggs:
hi @modax - can you show your use case where this breaks (i.e. - an example config)?

bryantbiggs:
I'm going to close this now as the current pattern is the desired state. I think the issue you are running into is a dependency conflict between when the outputs are available and when the provider can start using them

Issue #229, opened 2020-06-12
RDS Monitoring role should use region-specific AmazonRDSEnhancedMonitoringRole


OffCode:
For AWS accounts in China (e.g. Beijing) the ARN of AmazonRDSEnhancedMonitoringRole is slightly different: 
arn:**aws-cn**:iam::aws:policy/service-role/AmazonRDSEnhancedMonitoringRole

Currently the code is hard-wiring a value in db_instance module like the following:
```
resource "aws_iam_role_policy_attachment" "enhanced_monitoring" {
  count = var.create_monitoring_role ? 1 : 0

  role       = aws_iam_role.enhanced_monitoring[0].name
  policy_arn = "arn:aws:iam::aws:policy/service-role/AmazonRDSEnhancedMonitoringRole"
}
```

While creating a managed postgresql via terraform terraform-aws-rds module, I am getting the following error, which is cause due to using a non-existing policy from above:

```shell
fatal: [localhost]: FAILED! => {"changed": true, "cmd": "terraform apply -input=false tfplan -no-color", "delta": "0:00:23.741917", "end": "2020-06-12 06:46:35.446794", "msg": "non-zero return code", "rc": 1, "start": "2020-06-12 06:46:11.704877", "stderr": "\nError: Provider produced inconsistent result after apply\n\nWhen applying changes to\nmodule.db.module.db_instance.aws_iam_role_policy_attachment.enhanced_monitoring[0],\nprovider \"registry.terraform.io/-/aws\" produced an unexpected new value for\nwas present, but now absent.\n\nThis is a bug in the provider, which should be reported in the provider's own\nissue tracker.", "stderr_lines": ["", "Error: Provider produced inconsistent result after apply", "", "When applying changes to", "module.db.module.db_instance.aws_iam_role_policy_attachment.enhanced_monitoring[0],", "provider \"registry.terraform.io/-/aws\" produced an unexpected new value for", "was present, but now absent.", "", "This is a bug in the provider, which should be reported in the provider's own", "issue tracker."]
```

mehstg:
Also experiencing this issue.

cjpeterein:
This doesn't work for me when using the toplevel module, since that doesn't allow me to pass in the iam_partition variable.  Instead of passing in a variable, you could look up the partition directly:

data "aws_partition" "current" {}

PR #256, opened 2020-10-01
fix: Added possibility to specify partition used in iam role policy attachment

mehstg:
## Description
Currently the partition is hard coded when generating the ARN for the policy attachment. This will not work in China/Govcloud where the partition is not "aws"
## Motivation and Context
This issue fixes https://github.com/terraform-aws-modules/terraform-aws-rds/issues/229
We are experiencing the issue in the cn-north-1 region.

## Breaking Changes
No. Default value matches what is already in place.

## How Has This Been Tested?
This has been tested against our current infrastructure and works.


mehstg:
Possible to get this looked at?

antonbabenko:
Thanks, @mehstg !

[v2.20.0](https://github.com/terraform-aws-modules/terraform-aws-rds/releases/v2.20.0) has been just released.

PR #266, opened 2020-10-23
Don't create other resources if db instance is not created

robinbowes:
## Description
Don't create other resources if the DB instance is not created

## Motivation and Context
If I set `create_db_instance = false` I don't want the other resources to be created, eg. db_subnet_group, db_parameter_group, db_option_group

## Breaking Changes
I don't believe this breaks anything

## How Has This Been Tested?
Tested on Terraform Cloud with Terraform 0.12.29

### DIFF
--- a/README.md
+++ b/README.md
@@ -160,10 +160,10 @@ No provider.
 | ca\_cert\_identifier | Specifies the identifier of the CA certificate for the DB instance | `string` | `"rds-ca-2019"` | no |
 | character\_set\_name | (Optional) The character set name to use for DB encoding in Oracle instances. This can't be changed. See Oracle Character Sets Supported in Amazon RDS for more information | `string` | `""` | no |
 | copy\_tags\_to\_snapshot | On delete, copy all Instance tags to the final snapshot (if final\_snapshot\_identifier is specified) | `bool` | `false` | no |
-| create\_db\_instance | Whether to create a database instance | `bool` | `true` | no |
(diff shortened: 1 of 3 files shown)


robinbowes:
@antonbabenko How does this look?

antonbabenko:
Hi,

Users may want to create DB option/parameter groups even without a DB instance, so I am not sure this is a good idea to have such a strong dependency.

robinbowes:
> Hi,
> 
> Users may want to create DB option/parameter groups even without a DB instance, so I am not sure this is a good idea to have such a strong dependency.

Hi Anton,

In that case, they can just use the `aws_db_option_group`/`aws_db_parameter_group`  resource directly. There's no reason to use the rds module just to create the option/parameter groups - it adds no value, unless you're also creating the DB instance.

R.

antonbabenko:
You are right, there is more than one way to get job done - use root module, use submodule, and use resources directly.

Using resources directly is not great, because it is not possible to version resources or do any kind of logic inside of it. It is better in 0.12+ for sure.

Users are using the root module to get desired functionality and there is no reason for them to change their code to use submodule (and do `terraform state mv ...`).

I am going to close this issue.

robinbowes:
How about not ignoring the `create_db_option_group` setting? If the user sets `create_db_option_group = false` then don't create the DB option group. At the moment, that flag is ignored if the `option_group_name` is empty.

PR #269, opened 2020-11-18
fix: Ignore changes to snapshot_identifier

### LINKED ISSUE #270: RDS is destroyed on next terraform apply if snapshot_identifier is changed
If we restore an RDS instance from a snapshot then change the snapshot identifier, the RDS instance is deleted and re-created.

Some discussion of the issue here: https://discuss.hashicorp.com/t/rds-restored-from-snapshot-is-destroyed-on-next-terraform-apply/3693

robinbowes:
## Description
Ignore changes to `snapshot_identifier`. Fixes #270 

## Motivation and Context
`snapshot_identifier` is only used when creating the DB instance. However, if the parameter is changed post-creation it causes the DB instance to be deleted and re-created.

This is the solution recommended by Hashicorp: https://discuss.hashicorp.com/t/rds-restored-from-snapshot-is-destroyed-on-next-terraform-apply/3693/4

## Breaking Changes
This change should not break anything, unless a user is relying on changing this parameter to trigger the creation of a DB instance.

## How Has This Been Tested?
We are using this change in our private fork of this module.

### DIFF
--- a/modules/db_instance/main.tf
+++ b/modules/db_instance/main.tf
@@ -110,6 +110,10 @@ resource "aws_db_instance" "this" {
     delete = lookup(var.timeouts, "delete", null)
     update = lookup(var.timeouts, "update", null)
   }
+
+  lifecycle {
+    ignore_changes = [snapshot_identifier]
+  }
 }
 
 resource "aws_db_instance" "this_mssql" {
@@ -182,6 +186,10 @@ resource "aws_db_instance" "this_mssql" {
     create = lookup(var.timeouts, "create", null)
     delete = lookup(var.timeouts, "delete", null)
     update = lookup(var.timeouts, "update", null)
+  }
+
+  lifecycle {
+    ignore_changes = [snapshot_identifier]
   }
 }
 


robinbowes:
Close/re-open to re-trigger tests

robinbowes:
I don't understand why the "Semantic Pull Request" check is still failing. 🤷

grigorov (approved):

okgolove:
I'd suggest having two `aws_db_instance` resources. One with `ignore_changes`, one without.
In some cases I want to trigger recreating db instance based on snapshot identifier.

bryantbiggs:
hi @robinbowes - thanks for the PR. this is always something very tricky/difficult to work with in modules (lifecycle hooks). I think the safest approach is as @okgolove stated but that adds a significant amount of overhead and complexity for little gain - I'm not sure what we will be able to do here (hoping Hashicorp provide a variabilzed version of lifecycle hooks soon 🤞🏼)

bryantbiggs:
closing this in favor of what is provided by the latest release of the upstream AWS provider https://github.com/hashicorp/terraform-provider-aws/pull/18013

PR #277, opened 2021-01-18
fix: documentation for default storage type

saimonn:
## Description
I changed the doc to reflect reality on the **storage_type** variable.

## Motivation and Context
The documentation is somewhat confusing, the description explains it defaults to "standard", and the field "default" of the table reads "gp2".
We had to read the code to guess the real default storage_type (gp2).

## Breaking Changes
None.

## How Has This Been Tested?
I read the doc after the change: it was not confusing anymore.


bryantbiggs (approved):
@antonbabenko 👍🏼

PR #285, opened 2021-02-11
feat: Fetch credentials from AWS Secret Manager

senyo-asamoah-zefr:
## Description
Enable the db credentials to be fetched from the AWS Secret Manager

## Motivation and Context
Make creating and managing DB user credentials easy and secure since the credentials are stored and managed in a central place that can be rotated and lifecycle.

## Breaking Changes
YES
This change makes the assumption that you are on a version of `terraform >=0.12`, since makes use of the `jsondecode` function.
This change is important because it introduces the concept of having a centralized place for managing credentials and allows for immediate remediation of a breach by having anyone with admin access to the secrets manager rotate the keys in case of a breach

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -1,3 +1,13 @@
+# get the secret metadata
+data "aws_secretsmanager_secret" "db_creds" {
+  name = var.secret_name
+}
+
+#get the secret values using the id from the metadata
+data "aws_secretsmanager_secret_version" "db_creds_version" {
+  secret_id = data.aws_secretsmanager_secret.db_creds.id
+}
+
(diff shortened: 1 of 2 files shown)


justinrmiller:
👍🏻

bryantbiggs:
hi @senyo-asamoah-zefr - this looks great but I don't think it fits within the module. this would change the default behavior for all which may not be desirable. I think the best use case here would be to use the setup you have configured outside of the module and pass in your `local.creds_map.username` and `local.creds_map.password` into the module for `username` and `password`

PR #295, opened 2021-03-02
Adding Sensitive Property To Username

AMcManigal:
## Description

Terraform 14.7 will throw the following error.

```
Error: Output refers to sensitive values

  on .terraform/modules/db/modules/db_instance/outputs.tf line 75:
  75: output "this_db_instance_username" {

Expressions used in outputs can only refer to sensitive values if the
sensitive attribute is true.
```

## Motivation and Context
It allows Terraform 14.7 users to continue using the module.

## Breaking Changes
I don't think there should be any breaking changes.

## How Has This Been Tested?
I have tested this with a stack that uses this module. Before the change the previously mentioned error would keep me from running even a 'terraform plan'. After the change the module runs as expected.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,3 +27,6 @@ override.tf.json
 # Ignore CLI configuration files
 .terraformrc
 terraform.rc
+
+#Ignore Idea IDEs
+.idea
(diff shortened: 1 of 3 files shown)


antonbabenko:
The username is not sensitive.

PS: You submit to the wrong repo.

AdamFerguson:
After upgrading to terraform v0.15.0 I'm seeing this same error with this module. This module was working previous to the upgrade.

```
│ Error: Output refers to sensitive values
│ 
│   on .terraform/modules/rds/modules/db_instance/outputs.tf line 76:
│   76: output "this_db_instance_username" {
│ 
│ Expressions used in outputs can only refer to sensitive values if the sensitive attribute is true.
```

PR #298, opened 2021-03-07
chore: update example projects

bryantbiggs:
## Description
- update examples to use modules to setup supporting resources instead of relying on account level defaults which may not exist (i.e. - create a VPC to test vs trying to use default VPC which may not exist for some users)
- update examples to use later database versions to avoid running into any EOL issues
- validate all examples modified (except Oracle which does not include any changes at this time)

## Motivation and Context
- ensure that modules are working as intended and ready to be used to validate changes

## Breaking Changes
- no, only example changes, nothing that affects the module itself

## How Has This Been Tested?
- tested by executing all modules that have been changed to validate changes

### DIFF
--- a/examples/complete-mssql/main.tf
+++ b/examples/complete-mssql/main.tf
@@ -17,7 +17,7 @@ locals {
 
 module "vpc" {
   source  = "terraform-aws-modules/vpc/aws"
-  version = "~> 2.77"
+  version = "~> 2"
 
   name = local.name
   cidr = "10.99.0.0/18"
@@ -28,15 +28,13 @@ module "vpc" {
   database_subnets = ["10.99.7.0/24", "10.99.8.0/24", "10.99.9.0/24"]
 
   create_database_subnet_group = true
-  enable_nat_gateway           = true
-  single_nat_gateway           = true
 
   tags = local.tags
 }
 
 module "security_group" {
   source  = "terraform-aws-modules/security-group/aws"
-  version = "~> 3.18"
+  version = "~> 3"
 
   name        = local.name
   description = "Complete SqlServer example security group"
--- a/examples/complete-mssql/main.tf
+++ b/examples/complete-mssql/main.tf
@@ -118,9 +118,9 @@ module "db" {
   identifier = local.name
 
   engine               = "sqlserver-ex"
-  engine_version       = "14.00.1000.169.v1"
-  family               = "sqlserver-ex-14.0" # DB parameter group
-  major_engine_version = "14.00"             # DB option group
+  engine_version       = "15.00.4073.23.v1"
+  family               = "sqlserver-ex-15.0" # DB parameter group
+  major_engine_version = "15.00"             # DB option group
   instance_class       = "db.t3.large"
 
   allocated_storage     = 20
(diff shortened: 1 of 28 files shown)


examples/complete-mssql/main.tf
```diff
@@ -1,33 +1,64 @@
...
-data "aws_vpc" "default" {
-  default = true
-}
+################################################################################
+# Supporting Resources
+################################################################################
+
+module "vpc" {
+  source  = "terraform-aws-modules/vpc/aws"
+  version = "~> 2.77"
```
antonbabenko, line 20: I think that we can set `version` to be `~> 2`, so that we _also_ test compatibility of the whole range of versions and not just specific version 2.77. What do you think?

examples/complete-mssql/main.tf
```diff
@@ -1,33 +1,64 @@
...
+
+  azs              = ["${local.region}a", "${local.region}b", "${local.region}c"]
+  public_subnets   = ["10.99.0.0/24", "10.99.1.0/24", "10.99.2.0/24"]
+  private_subnets  = ["10.99.3.0/24", "10.99.4.0/24", "10.99.5.0/24"]
+  database_subnets = ["10.99.7.0/24", "10.99.8.0/24", "10.99.9.0/24"]
 
-data "aws_subnet_ids" "all" {
-  vpc_id = data.aws_vpc.default.id
+  create_database_subnet_group = true
+  enable_nat_gateway           = true
```
antonbabenko, line 31: Please disable NAT gateway to save time during runs.

examples/complete-mysql/main.tf
```diff
@@ -1,75 +1,101 @@
...
-  allocated_storage = 5
-  storage_encrypted = false
-
-  # kms_key_id        = "arm:aws:kms:<region>:<account id>:key/<kms key id>"
-  name     = "demodb"
-  username = "user"
-  password = "YourPwdShouldBeLongAndSecure!"
-  port     = "3306"
+  create_database_subnet_group = true
+  enable_nat_gateway           = true
```
antonbabenko, line 31: Same with NAT here and in all other examples.

examples/complete-mssql/main.tf
```diff
@@ -1,33 +1,64 @@
...
+
+  tags = local.tags
 }
 
-data "aws_security_group" "default" {
-  vpc_id = data.aws_vpc.default.id
-  name   = "default"
+module "security_group" {
+  source  = "terraform-aws-modules/security-group/aws"
+  version = "~> 3.18"
```
antonbabenko, line 39: Also with this version range. Let's use `~> 3`. Same on all other examples.

examples/s3-import-mysql/main.tf
```diff
@@ -85,7 +85,7 @@ module "security_group" {
 
 module "import_s3_bucket" {
   source  = "terraform-aws-modules/s3-bucket/aws"
-  version = "1.17.0"
+  version = "~> 1.20"
```
antonbabenko, line 88: `~> 1` :)

antonbabenko (changes requested):
Minor comments about version ranges.

examples/complete-mssql/main.tf
bryantbiggs: sounds good - updated!

examples/complete-mssql/main.tf
bryantbiggs: Done

examples/complete-mysql/main.tf
bryantbiggs: Done

examples/complete-mssql/main.tf
bryantbiggs: Done

examples/s3-import-mysql/main.tf
bryantbiggs: Done

antonbabenko:
Thanks, [v2.23.0](https://github.com/terraform-aws-modules/terraform-aws-rds/releases/v2.23.0) has been just released.

PR #282, opened 2021-02-06
fix: remove db_option_group for postgres engine, as not in use

kuzm1ch:
## Description
* avoid db_option_group creating in case DB engine is postgres 
* minor update example based on it

## Motivation and Context
```
PostgreSQL does not use options and option groups. PostgreSQL uses extensions and modules to provide additional features. For more information, see Supported PostgreSQL features and extensions.
```
https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/USER_WorkingWithOptionGroups.html 

## Breaking Changes
In case "PostgresSQL option" was previosly created, it could some problem while removing it with next apply(cannot be deleted because it is in use). 
`terrafor state rm` can be used as a workaround.
```
terraform state rm module.dbnew.module.db_option_group.aws_db_option_group.this
```

## How Has This Been Tested?
examples/complete-postgres/main.tf

PR #302, opened 2021-03-08
fix: update option group to fix name vs. name_prefix and conditional creation for postgresql

### LINKED ISSUE #188: DB option group is created with create_db_option_group = false parameter
Hi,

Currently it is not possible to avoid creating a DB option group only with a workaround (setting `option_group_name` to some dummy value): 
https://github.com/terraform-aws-modules/terraform-aws-rds/blob/3997fcb615d1fa516cc7815e12b130a9cd987f3c/main.tf#L9

Was this really the intention?

Thanks!

### LINKED ISSUE #230: Option group always with name_prefix and timestamp generated by terraform
"name" variable, in the option group module,  is missing.
The option group name is always a concatenation of "name_prefix" and and a timestamp generated by terraform.
"name" variable may be added in the option group module and, if set, it can be used to give a name without timestamp to the option group.
Thanks

### LINKED ISSUE #272: Option Group gets created even if key is set to false
When setting creation of replica to false, the DB option group for replica still gets created.

tfvars
```
db_create = true
db_replica_create = false
```

tf
```
module "db" {
    source = "terraform-aws-modules/rds/aws"
    version= "~> 2.20"

    create_db_instance        = var.db_create
    create_db_option_group    = var.db_create
    create_db_parameter_group = var.db_create
    create_db_subnet_group    = var.db_create
    ...
}

module "db_replica" {
    source = "terraform-aws-modules/rds/aws"
    version= "~> 2.20"

    create_db_instance        = var.db_replica_create
    # replicas do not have these configurations
    create_db_option_group    = false
    create_db_parameter_group = false
    create_db_subnet_group    = false

    replicate_source_db = module.db.this_db_instance_id
    ...
}
```

TF output
```
module.db_replica.module.db_option_group.aws_db_option_group.this[0]: Creating...
```

bryantbiggs:
## Description
- update option group `name` to use `option_group_name` provided by root module
- update option group to allow users to select to use `name_prefix` or `name` depending on value of `option_group_use_name_prefix`
- when `engine == 'postgresql'`, no option group is created and a null value provided to the root module ([PostgreSQL does not use option groups](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/USER_WorkingWithOptionGroups.html))
- added an example project to verify/validate changes and also updated the `README.md` to show different configuration combinations for users

## Motivation and Context
- give users the ability to better control the management of option groups and to avoid having to set values for PostgreSQL where option groups are not utilized

Closes #188 
Closes #230
Closes #251 
Closes #259 
Closes #272 
Closes #282

## Breaking Changes
- Yes but albeit minor. We should add a note about upgraded to show users how they can upgrade without any infra changes (though, its just an option group so its quite safe to re-create):

Users who create an option group using the module should change the following to avoid any re-creation:
```hcl
	# Add
	option_group_name = "${var.identifier}-"
``` 

## How Has This Been Tested?
- Tested using the provided `examples/option-groups` example directory

### DIFF
--- a/README.md
+++ b/README.md
@@ -114,6 +114,39 @@ module "db" {
 }
 ```
 
+## Option Groups
+
+[Reference](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/USER_WorkingWithOptionGroups.html)
+
+Users have the ability to:
+
+- Create an option group with the name provided:
+
+```hcl
+  option_group_name            = "prod-instance-mysql-8.0"
+  option_group_use_name_prefix = false
+```
+
+- Create an option group using a unique prefix beginning with the name provided:
+
+```hcl
+  option_group_name = "prod-instance-mysql-8.0"
+```
+
+- Pass the name of an option group to use that has been created outside of the module:
+
+```hcl
+  create_option_group = false
+  option_group_name   = "prod-instance-mysql-8.0" # must already exist in AWS
+```
+
+- Skip creating an option group for PostgreSQL entirely as that is not supported
+
+```hcl
+  engine            = "postgres"
+  option_group_name = "prod-instance-postgresql-11.0" # this will be ignored, no option group created
+```
+
 ## Examples
 
 - [Complete RDS example for MSSQL](https://github.com/terraform-aws-modules/terraform-aws-rds/tree/master/examples/complete-mssql)
@@ -198,8 +231,9 @@ No resources.
 | multi\_az | Specifies if the RDS instance is multi-AZ | `bool` | `false` | no |
 | name | The DB name to create. If omitted, no database is created initially | `string` | `""` | no |
 | option\_group\_description | The description of the option group | `string` | `""` | no |
-| option\_group\_name | Name of the DB option group to associate | `string` | `""` | no |
+| option\_group\_name | Name of the option group | `string` | `""` | no |
 | option\_group\_timeouts | Define maximum timeout for deletion of `aws_db_option_group` resource | `map(string)` | <pre>{<br>  "delete": "15m"<br>}</pre> | no |
+| option\_group\_use\_name\_prefix | Determines whether to use `option_group_name` as is or create a unique name beginning with the `option_group_name` as the prefix | `bool` | `true` | no |
 | options | A list of Options to apply. | `any` | `[]` | no |
 | parameter\_group\_description | Description of the DB parameter group to create | `string` | `""` | no |
 | parameter\_group\_name | Name of the DB parameter group to associate or create | `string` | `""` | no |
--- a/examples/option-groups/README.md
+++ b/examples/option-groups/README.md
@@ -17,7 +17,10 @@ Note that this example may create resources which cost money. Run `terraform des
 <!-- BEGINNING OF PRE-COMMIT-TERRAFORM DOCS HOOK -->
 ## Requirements
 
-No requirements.
+| Name | Version |
+|------|---------|
+| terraform | >= 0.12.26 |
+| aws | >= 2.49 |
 
 ## Providers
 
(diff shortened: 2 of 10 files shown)


antonbabenko:
Looks good to me once CI is happy again.

bryantbiggs:
ya, looks like I forgot the version block in the new example, should be all set for review now @antonbabenko

antonbabenko:
Awesome work! 🚀 

[v2.24.0](https://github.com/terraform-aws-modules/terraform-aws-rds/releases/v2.24.0) has been just released.

paul-pop:
I'm afraid this has now broken my Terraform run (I'm using Postgres and I have an instance running with an option group - although not supported, it still created it a long time ago).

Unfortunately, there are cases where the option group can't just be removed immediately from a Postgres instance because of snapshots being attached to the option group. Here is the plan I got:

```hcl
# module.rds_app_db.module.db_instance.aws_db_instance.this[0] will be updated in-place
  ~ resource "aws_db_instance" "this" {
        id                                    = "my-dev-app-db"
        name                                  = "my"
      ~ parameter_group_name                  = "my-dev-app-db-20201027145803537200000001" -> (known after apply)

        # (48 unchanged attributes hidden)

        # (1 unchanged block hidden)
    }

  # module.rds_app_db.module.db_option_group.aws_db_option_group.this[0] will be destroyed
  - resource "aws_db_option_group" "this" {
      - arn                      = "arn:aws:rds:eu-west-2:xxxx:og:my-dev-app-db-20201027145803542000000002" -> null
      - engine_name              = "postgres" -> null
      - id                       = "my-dev-app-db-20201027145803542000000002" -> null
      - major_engine_version     = "12" -> null
      - name                     = "my-dev-app-db-20201027145803542000000002" -> null
      - name_prefix              = "my-dev-app-db-" -> null
      - option_group_description = "Option group for my-dev-app-db" -> null

      - timeouts {
          - delete = "15m" -> null
        }
    }

  # module.rds_app_db.module.db_parameter_group.aws_db_parameter_group.this[0] must be replaced
+/- resource "aws_db_parameter_group" "this" {
      ~ arn         = "arn:aws:rds:eu-west-2:xxx:pg:my-dev-app-db-20201027145803537200000001" -> (known after apply)
      ~ description = "Database parameter group for my-dev-app-db" -> "my-dev-app-db parameter group" # forces replacement
      ~ id          = "my-dev-app-db-20201027145803537200000001" -> (known after apply)
      ~ name        = "my-dev-app-db-20201027145803537200000001" -> (known after apply)

        # (2 unchanged attributes hidden)

        # (4 unchanged blocks hidden)
    }
```

So if you have an RDS cluster that performs automated backups, you would need to manually remove all snapshots (which is also not great because they can only be removed by reducing the backup window to 0). The error message you get when attempting to delete:

```
You can’t delete a system snapshot. To remove system snapshots for an active DB instance, modify the DB instance and reduce the automated backup window. To remove system snapshots for a deleted instance, delete the retained automated backup.
```
 
My workaround was to:

1. Set backup window to 0
2. I used the AWS CLI to update to the default option group: 
```
aws rds modify-db-instance --db-instance-identifier my-dev-app-db --option-group-name default:postgres-12 --apply-immediately
```
3. Wait ~5 mins for RDS instance to be modified
4. Terraform apply - which will remove the option group and re-enable automated backups

@bryantbiggs @antonbabenko

bryantbiggs:
@paul-pop while I am deeply sorry that this has caused issues on your end, and potentially others, it is the desired effect that we want for the module. I am not sure why the AWS API allows for the creation of an option group for PostgreSQL instances while stating that its not supported in the docs, we are trying to match those same intentions here in the module

paul-pop:
@bryantbiggs I completely understand and this is indeed a bug on the AWS side because as soon as you create an option group for Postgres it hangs when you want to add an option so this module avoids getting into that.

Luckily for us, we're ok to lose the snapshots but other people might find it hard to do so. Probably worth keeping an eye out for issues that are raised because of this and suggesting my workaround or something similar.

bryantbiggs:
yes, and thank you for sharing very detailed steps @paul-pop - I am sure it will be quite helpful for others who will come across this issue

PR #304, opened 2021-03-08
fix: update parameter group to fix name vs. name_prefix

bryantbiggs:
## Description
- remove 2nd resource in `parameter_group` sub-module and instead use a variable flag to determine whether or not to use a unique prefix on the name; this matches what was done in the `option_group` sub-module
- remove use of `var.identifier` from sub-module and instead rely on `var.parameter_group_name` provided 
- update `example/option-groups` -> `option/groups` to test/validate changes to option and parameter groups outside of creating RDS instances

## Motivation and Context
- working towards consistent use of `name` versus `name_prefix` for the sub-modules as well as the root module (subnet group will be next and last update required now that option group is done + this PR)

Relates to #121

## Breaking Changes
yes; albeit minor since again, parameter groups can be re-created without disruption (`create_before_destroy` is used here)

For users who are looking to upgrade, they will need to perform the following changes if they wish to forgo the re-creation:

If using without the name-prefix for parameter groups:
- `terraform state mv module.<your_module_name>.db_parameter_group.this_no_prefix module.<your_module_name>.db_parameter_group.this`

## How Has This Been Tested?
- Tested and validated using the update `examples/groups` directory


atrakic:
I didn't have much time to see why it failed as CI failed all of sudden so as quick fix, Ive ended up to pin `version = "2.23.0"`. Hope it helps to somebody.

atrakic:
BTW: Just tried to bump to latest version with ```create_db_option_group          = false``` but after removing state: 

```
terraform state rm "module.rds.module.db_option_group.aws_db_option_group.this[0]"
```
my run ends with:  
```
Error: Error in function call

  on .terraform/modules/rds/main.tf line 8, in locals:
   8:   option_group           = var.engine != "postgres" ? coalesce(module.db_option_group.this_db_option_group_id, var.option_group_name) : null
    |----------------
    | module.db_option_group.this_db_option_group_id is ""
    | var.option_group_name is ""

Call to function "coalesce" failed: no non-null, non-empty-string arguments.


Error: first character of parameter group "name_prefix" must be a letter

  on .terraform/modules/rds/modules/db_parameter_group/main.tf line 12, in resource "aws_db_parameter_group" "this":
  12:   name_prefix = local.name_prefix

```
Comment: coalesce() supports default values, maybe should be used here?

bryantbiggs:
hi @atrakic, I have opened a PR that should do two things that have been identified:
- if you do not provide a name for the option group or parameter group, it will default back to the previous default of `${var.identifier}-`
- if you elect to not create an option group or parameter group, nor provide a name of an existing option/parameter group - it will default to pulling the default groups provided by AWS

I've updated the two examples to show and validate this 
https://github.com/terraform-aws-modules/terraform-aws-rds/pull/308

atrakic:
@bryantbiggs Yes, after second attempt I was able to bump to latest with:
```
...
create_db_subnet_group = true
  parameters = [
    {
      name  = "character_set_client"
      value = "utf8"
    },
    {
      name  = "character_set_server"
      value = "utf8"
    }
  ]

  options = [
    {
      option_name = "MARIADB_AUDIT_PLUGIN"

      option_settings = [
        {
          name  = "SERVER_AUDIT_EVENTS"
          value = "CONNECT"
        },
        {
          name  = "SERVER_AUDIT_FILE_ROTATIONS"
          value = "37"
        },
      ]
    },
  ]
```

PR #307, opened 2021-03-09
fix: only create option group if create_db_option_group is set

bondezbond:
## Description
This will fix where we can still use default option_parameter from AWS Provider

## Motivation and Context
* Fix for pattern that was established before release v2.25.0

## Breaking Changes
None found

## How Has This Been Tested?
There should be no breaking changes. I tested with a local project I was using before hand


bondezbond:
closing in favor of #308

PR #310, opened 2021-03-10
fix: add backward compatible for the optional group for PostgreSQL

### LINKED ISSUE #309: Backward compatible for skip creating the optional group for PostgreSQL
## Description

Version v2.24.0  contains changes that are not backward compatible. If the optional group for PostgreSQL was created earlier, a new version will require it to be removed. Removing the optional group will require additional steps.

## How to reproduce
- Create RDS PostgreSQL with the optional group using version v2.23.0 (or previous)
- Update module to version v2.24.0 (or newer)
- Run 'terraform apply'
```
  # module.db_default.module.db_option_group.aws_db_option_group.this[0] will be destroyed
  - resource "aws_db_option_group" "this" {
      - arn                      = "arn:aws:rds:eu-central-1:108859834221:og:complete-postgresql-default-20210310182549840800000003" -> null
      - engine_name              = "postgres" -> null
      - id                       = "complete-postgresql-default-20210310182549840800000003" -> null
      - major_engine_version     = "11" -> null
      - name                     = "complete-postgresql-default-20210310182549840800000003" -> null
      - name_prefix              = "complete-postgresql-default-" -> null
      - option_group_description = "complete-postgresql-default option group" -> null
      - tags                     = {
          - "Environment" = "dev"
          - "Name"        = "complete-postgresql-default"
          - "Owner"       = "user"
        } -> null

      - timeouts {
          - delete = "15m" -> null
        }
    }

Plan: 0 to add, 0 to change, 1 to destroy.
```
- Waiting for error
```
Error: Error Deleting DB Option Group: InvalidOptionGroupStateFault: The option group 'complete-postgresql-default-20210310182549840800000003' cannot be deleted because it is in use.
        status code: 400, request id: 13a154ab-77cb-4eae-bdf0-470963822a91
```
- Try from AWS Managment Console
Selection_20210311_0001
Selection_20210311_0002

cageyv:
## Description
Backward compatible for the optional group for PostgreSQL
An additional optional variable that allows us to leave db_option_group for PostgreSQL in place and not delete it.
It will come in handy if it was previously created.

## Motivation and Context
It is not always possible to immediately remove db_option_group
I propose to add the ability to tackle this problem separately and continue to use the new functions of the module
Close #309 

## Breaking Changes
No

## How Has This Been Tested?
Tested with `complete-postgres` example

### DIFF
--- a/README.md
+++ b/README.md
@@ -147,6 +147,15 @@ Users have the ability to:
   option_group_name = "prod-instance-postgresql-11.0" # this will be ignored, no option group created
 ```
 
+- Backward compatibility in case the optional group for PostgreSQL was created earlier
+
+```hcl
+  engine            = "postgres"
+  option_group_name = "prod-instance-postgresql-11.0" # this will be used, option group created
+
+  skip_db_option_group_for_postgres = false
+```
+
 ## Examples
 
 - [Complete RDS example for MSSQL](https://github.com/terraform-aws-modules/terraform-aws-rds/tree/master/examples/complete-mssql)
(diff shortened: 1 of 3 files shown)


tovbinm:
Can anyone review this please?

antonbabenko:
HI @cageyv !

I don't think we need to have an extra condition and another variable for the case.

If the user doesn't want to get resource created, set `create_db_option_group = false`.

If after some migration or anything the user wants to not manage some resource via Terraform, they can run `terraform state rm ...`.

I don't think this PR is something we really need to have merged.

@bryantbiggs what do you think?

PR #311, opened 2021-03-11
fix: Adds missing ARN for the monitoring role

danvaida:
## Description

This PR address the issue of the missing `monitoring_role_arn` parameter for the `aws_db_instance` resource, in case of MS SQL.

## Motivation and Context
Getting the module to work as advertised.

## Breaking Changes
N/A

## How Has This Been Tested?

Terraform v0.13.5
Module version: 2.18.0

### DIFF
--- a/modules/db_instance/main.tf
+++ b/modules/db_instance/main.tf
@@ -155,12 +155,12 @@ resource "aws_db_instance" "this_mssql" {
   parameter_group_name   = var.parameter_group_name
   option_group_name      = var.option_group_name
 
-  availability_zone   = var.availability_zone
-  multi_az            = var.multi_az
-  iops                = var.iops
-  publicly_accessible = var.publicly_accessible
-  monitoring_interval = var.monitoring_interval
-  monitoring_role_arn = var.monitoring_interval > 0 ? coalesce(var.monitoring_role_arn, join(", ", aws_iam_role.enhanced_monitoring.*.arn), null) : null
+  availability_zone           = var.availability_zone
+  multi_az                    = var.multi_az
+  iops                        = var.iops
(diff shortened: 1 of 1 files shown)


bryantbiggs:
hi @danvaida - what is the issue that you are facing? have you tried the latest module version?

Nickmman:
@bryantbiggs Facing the same issue. It's due to v2.29.0, the latest module version does not seem to have a fix for this yet.

Issue is as follows:
```
Error: Incorrect attribute value type
  on .terraform/modules/db/modules/db_instance/main.tf line 157, in resource "aws_db_instance" "this_mssql":
 157:   monitoring_role_arn     = var.monitoring_interval > 0 ? coalesce(var.monitoring_role_arn, aws_iam_role.enhanced_monitoring.*.arn, null) : null
    |----------------
    | aws_iam_role.enhanced_monitoring is empty tuple
    | var.monitoring_interval is 0
    | var.monitoring_role_arn is null
Inappropriate value for attribute "monitoring_role_arn": string required.
```

Mind you that on our end, the module was working as-is (prior to v2.29.0) for 11+ months. The issue that we're experiencing is also for MSSQL.

oirtemed:
Hi @bryantbiggs ,

We're also experiencing the same issue.

The code change proposed in this PR seems to fix it.

Any idea when this fix can be merged?

bryantbiggs:
hi all, not sure what happened there but you are correct, there is an issue with the monitoring role ARN. this PR won't do much since it matches what is already on `master`, but I have opened a PR #318 to address this

antonbabenko:
Please use [v2.33.0](https://github.com/terraform-aws-modules/terraform-aws-rds/releases/v2.33.0) which has been just released.

PR #315, opened 2021-03-15
chore: remove default values and replace with `null` to use upstream default settings

### LINKED ISSUE #235: RDS db_instance storage type default
https://github.com/terraform-aws-modules/terraform-aws-rds/tree/master/modules/db_instance

The storage type should default to 'gp2' not 'standard'. This has been updated in RDS documentation as well.

### LINKED ISSUE #273: Postgres: apply fails due to use of Oracle character_set_name
For some reason this only occurs on some of our instances, but has a clear cause, and is triggering terraform 0.12+ trying to prevent the formation of inconsistent plans:

https://www.terraform.io/docs/extend/terraform-0.12-compatibility.html#inaccurate-plans

bryantbiggs:
## Description
- remove default values and replace with `null` to use upstream default settings

## Motivation and Context
- the default values provided by the AWS provider/API will change periodically. instead of hard-coding those values in this module as the module default values, switching to `null` will allow the module to be flexible and change with the upstream source as needed

Closes #235
Closes #273 

## Breaking Changes
- No

## How Has This Been Tested?
- tested using `complete-postgresql` - no diff change between `master` and this PR


antonbabenko:
[v2.30.0](https://github.com/terraform-aws-modules/terraform-aws-rds/releases/v2.30.0) has been just released.

PR #321, opened 2021-03-23
feat: Add support for legacy parameter `security_group_names`

y-batsianouski:
## Description

Add support for `security_group_names` legacy parameter

## Motivation and Context

Even it's deprecated, but still required when you need to import existing non-vpc RDS instance (our case)

## Breaking Changes
-

## How Has This Been Tested?
It works on our infrastructure for both vpc RDS instances and non-vpc legacy RDS instances.


bryantbiggs:
with EC2 classic retiring https://aws.amazon.com/blogs/aws/ec2-classic-is-retiring-heres-how-to-prepare/ - I don't think this addition would be warranted

Issue #328, opened 2021-04-21
Missing sensitive values for outputs when upgrading to Terraform 0.15


dynamike:
## Description

We are in the process of testing out Terraform 0.15 for some of our services and came across an error with this module. 

## Versions

 - Terraform: 0.15

 - Provider(s): 
   - provider registry.terraform.io/hashicorp/aws v3.37.0
   - provider registry.terraform.io/hashicorp/random v3.1.0

 - Module: 2.34.0

## Reproduction

1. Run `terraform plan` on Terraform 0.14.x with an existing instance of this module with 0.14 and ensure no changes.
2. Upgrade local Terraform CLI to 0.15.0 
3. Run `terraform plan` and get the following error
```
╷
│ Error: Output refers to sensitive values
│
│ on .terraform/modules/db/modules/db_instance/outputs.tf line 71:
│ 71: output "this_db_instance_name" {
│
│ Expressions used in outputs can only refer to sensitive values if the sensitive attribute is true.
╵
╷
│ Error: Output refers to sensitive values
│
│ on .terraform/modules/db/modules/db_instance/outputs.tf line 76:
│ 76: output "this_db_instance_username" {
│
│ Expressions used in outputs can only refer to sensitive values if the sensitive attribute is true.
```

## Expected behavior
No changes in Terraform plan 

## Actual behavior
Error listed above around sensitive values associated with outputs.

## Additional context
I suspect this relates fo some new changes around sensitive outputs in Terraform 0.15 . See https://www.terraform.io/upgrade-guides/0-15.html#sensitive-output-values. I'll likely circle back to this and submit a fix soon if no one else gets to it first.

bryantbiggs:
hmm, this looks like either route we go is a breaking change:
1. explicitly set the output as sensitive from the module
2. use the `nonsensitive()` function which is only available from v0.14 https://www.terraform.io/docs/language/functions/nonsensitive.html

I suspect we'll have to probably look at option 1 as its less of an impact - other convos about this are happening over at https://github.com/terraform-aws-modules/terraform-aws-rds-aurora/issues/212

@antonbabenko any thoughts?

antonbabenko:
Definitely option 1. Less impact is better. We are still increasing the version requirements to Terraform 0.13 across all modules and I don't think we are ready to think about 0.14 features in the modules just yet. :)

[...]

PR #331, opened 2021-04-27
added a sensitive attribute to db instance name...

jerels:
## Description
added a sensitive attribute to db_instance_name in the outputs terraform db instance module, and set it to true

## Motivation and Context
While attempting to update to the latest version of the terraform-aws-rds module, ran into the following error:
` Error: Output refers to sensitive values
│ 
│   on .terraform/modules/db/modules/db_instance/outputs.tf line 71:
│   71: output "this_db_instance_name" {
│ 
│ Expressions used in outputs can only refer to sensitive values if the
│ sensitive attribute is true.`

A similar bug a colleague of mine had: https://github.com/terraform-aws-modules/terraform-aws-rds/issues/328

So I just took the fix from that issue and applied it to the db_instance_name in output.tf

### DIFF
--- a/modules/db_instance/outputs.tf
+++ b/modules/db_instance/outputs.tf
@@ -71,6 +71,7 @@ output "db_instance_status" {
 output "db_instance_name" {
   description = "The database name"
   value       = local.db_instance_name
+  sensitive   = true
 }
 
 output "db_instance_username" {


bryantbiggs (approved):
@antonbabenko 👍🏼

antonbabenko:
@bryantbiggs Maybe you can explain why `db_instance_name` is sensitive?

I don't understand why some arguments (like instance name and username) can be marked as sensitive? Is this a bug somewhere?

I just don't want to blindly add extra work for us to add and then to remove this sensitive feature from the modules :)

jerels:
thanks for the help!! but updating to the latest patch actually fixed our problem!!

Issue #316, opened 2021-03-15
Regression with db_subnet_group for replicas


JanKoppe:
## Description
Hello,

it seems like 1245fac0c34e09e3145fa0f1a6084c9465097273 broke working with DB replicas. I'm getting the following error even when trying to run the provided `examples/replica-mysql`:

```
terraform-aws-rds/examples/replica-mysql on  master took 3s
➜ terraform apply

Error: Error in function call

  on ../../main.tf line 2, in locals:
   2:   db_subnet_group_name = coalesce(var.db_subnet_group_name, module.db_subnet_group.this_db_subnet_group_id)
    |----------------
    | module.db_subnet_group.this_db_subnet_group_id is ""
    | var.db_subnet_group_name is null

Call to function "coalesce" failed: no non-null, non-empty-string arguments.

```
As far as I understand the docs (and it's even mentioned in some comments) replicas in the same region should not have a subnet group defined on the resource. The recent change assumes that a subnet group must always be defined, though.

## Versions

 - Terraform:
Terraform v0.14.8
+ provider registry.terraform.io/hashicorp/aws v3.32.0
+ provider registry.terraform.io/hashicorp/random v3.1.0
 - Provider(s):
Terraform v0.14.8
+ provider registry.terraform.io/hashicorp/aws v3.32.0
+ provider registry.terraform.io/hashicorp/random v3.1.0
 - Module:

## Reproduction
Checkout the repository, and run the `examples/replica-mysql` code.

## Expected behavior
Replicas should work without any subnet group defined.

## Actual behavior
The code requires either a db_subnet_group_name to be defined, or a db subnet group to be created with the db_subnet_group submodule.

davidkam:
Temporary workaround, add this to the replica module block:
`db_subnet_group_name = module.master.this_db_subnet_group_id`

antonbabenko:
Please use [v2.31.0](https://github.com/terraform-aws-modules/terraform-aws-rds/releases/v2.31.0) which has been just released.

PR #335, opened 2021-05-13
fix: Do not coalesce db_subnet_group_name when var.create_db_subnet_group == false

evanstachowiak:
Fixes this error:
```
    | module.db_subnet_group.db_subnet_group_id is ""
    | var.db_subnet_group_name is null

Call to function "coalesce" failed: no non-null, non-empty-string arguments.
```

Occurs when var.create_db_subnet_group == false.

Hopefully this fixes these regression issues that keep popping up, e.g. https://github.com/terraform-aws-modules/terraform-aws-rds/issues/316

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -1,6 +1,6 @@
 locals {
   master_password      = var.create_db_instance && var.create_random_password ? random_password.master_password[0].result : var.password
-  db_subnet_group_name = var.replicate_source_db != null ? null : coalesce(var.db_subnet_group_name, module.db_subnet_group.db_subnet_group_id)
+  db_subnet_group_name = var.replicate_source_db != null || var.create_db_subnet_group == false ? null : coalesce(var.db_subnet_group_name, module.db_subnet_group.db_subnet_group_id)
 
   parameter_group_name_id = var.create_db_parameter_group ? module.db_parameter_group.db_parameter_group_id : var.parameter_group_name
 


bryantbiggs:
are you setting `db_subnet_group_name` if you are setting `create_db_subnet_group = false`? https://github.com/terraform-aws-modules/terraform-aws-rds/blob/27c6fa5c24bac2d841c29b8160a725c91c47ffcc/examples/replica-postgres/main.tf#L92

evanstachowiak:
hi @bryantbiggs, thanks for your quick reply.

No, I am not setting `db_subnet_group_name`. Nor am I creating an RDS instance, I am setting `create_db_instance = false`.

I am using this module in the context where an RDS instance may or may not be created.

thpang:
@evanstachowiak Modified line number 3 just a bit:

```
db_subnet_group_name = var.replicate_source_db != null ? null : var.create_db_subnet_group == false ? null : coalesce(var.db_subnet_group_name, module.db_subnet_group.db_subnet_group_id)
```
Moved the flag check on the other side of the replicate_source_db logic. Testing now but seems promising.

main.tf
```diff
@@ -1,6 +1,6 @@
 locals {
   master_password      = var.create_db_instance && var.create_random_password ? random_password.master_password[0].result : var.password
-  db_subnet_group_name = var.replicate_source_db != null ? null : coalesce(var.db_subnet_group_name, module.db_subnet_group.db_subnet_group_id)
+  db_subnet_group_name = var.replicate_source_db != null || var.create_db_subnet_group == false ? null : coalesce(var.db_subnet_group_name, module.db_subnet_group.db_subnet_group_id)
```
thpang, line 3: Change this value to this:
```
db_subnet_group_name = var.replicate_source_db != null ? null : var.create_db_subnet_group == false ? null : coalesce(var.db_subnet_group_name, module.db_subnet_group.db_subnet_group_id)
```
This works in honoring new logic and old flag logic. Testing now.

evanstachowiak:
thanks everyone!

PR #338, opened 2021-05-19
feat: Add restore_to_point_in_time support for databases

luedigernet:
## Description
This adds restore_to_point_in_time support for databases others than mysql.

## Motivation and Context
When you have to perform a point in time recovery on a AWS RDS Instance it will always create a new instance.
In order to get this new instance under control of terraform I would use the same terraform-aws-module that was used to create the source database. The existing module is using the  aws_db_instance resource of the aws provider. Because this resource supports already the restore_to+point_in_time Input this module should support it as well.

## Breaking Changes
No braking changes.

## How Has This Been Tested?
We tested it with the following steps
- creating an AWS  RDS instance with this module
- creating a table and inserted some data with actual timestamp of inserting.
- waiting until the AWS console reported a latest available timestamp for recovery after our inserted data
- created a new instance with the restore_to_point_in_time option using the identifier of the source database and a timestamp in the middle of 2 inserted rows in the source database
- checking after finished instance creation that the data up to the provided timestamp exists and the data behind it does not exist.


modules/db_instance/main.tf
```diff
@@ -78,6 +78,19 @@ resource "aws_db_instance" "this" {
...
 
+
+  dynamic "restore_to_point_in_time" {
+
+    for_each = var.restore_to_point_in_time != null ? [var.restore_to_point_in_time] : []
+    content {
+      restore_time                  = restore_to_point_in_time.value.restore_time
+      source_db_instance_identifier = lookup(restore_to_point_in_time.value, "source_db_instance_identifier", null)
+      source_dbi_resource_id        = lookup(restore_to_point_in_time.value, "source_dbi_resource_id", null)
+      use_latest_restorable_time    = lookup(restore_to_point_in_time.value, "use_latest_restorable+time", null)
```
khdevel, line 89: There is a typo here - we have `use_latest_restorable+time` but should be `use_latest_restorable_time`

khdevel:
How do you see this implementation? When will it be available in standard release? @antonbabenko

modules/db_instance/main.tf
antonbabenko: fixed

antonbabenko:
[v3.2.0](https://github.com/terraform-aws-modules/terraform-aws-rds/releases/v3.2.0) has been just released.

PR #339, opened 2021-05-21
chore: Update README.md to current version

bmalinconico:
## Description
I spent a good 20 minutes trying to figure out the correct output names. Turns out the docs show the output names from V3 but the example to copy is pinning to 2.X

## Motivation and Context
Readme isn't going to help users who are new to the module.

## Breaking Changes
No

## How Has This Been Tested?
I can update some if needed, but the extent of testing was that it made my build green.


antonbabenko:
Thanks for the update!

PR #346, opened 2021-06-28
feat: allow the creation of cross-region rds replicas

riccardo-salamanna:
## Description
The changes adds a boolean called cross_region_replica.

## Motivation and Context
In order to create a cross-region replica, we should specify a target subnet group, which is not needed for a same-region replica, as your module already supports with no issues. The idea is to set the cross_region_replica variable to true to allow the creation of the needed subnet group with minor changes to the source code

## Breaking Changes
none

## How Has This Been Tested?
i've tested this on both a same region and cross region replica.


oscar-munoz (approved):

antonbabenko:
Hi!

How about changing the variable name from single-purpose `var.cross_region_replica` to a more general like - `var.create_db_subnet_group` with `default = true` (as it is now) ?

The reason for this change is that there can be other situations where it makes sense to disable the creation of db_subnet_group in the future (if not already).

riccardo-salamanna:
Hi @antonbabenko thanks for your quick reply. I might be misunderstanding but `var.create_db_subnet_group` is already `default = true`, but the modules as they are right now will always pass a string to the resource `(db_subnet_group_name   = local.db_subnet_group_name),` triggering the api error `Error: Error creating DB Instance: DBSubnetGroupNotAllowedFault: DbSubnetGroupName should not be specified for read replicas that are created in the same region as the master`
But i am totally opened to any solution from the maintainers that would allow us to create cross-region replicas, I submit to your expertise

main.tf
```diff
@@ -1,6 +1,6 @@
 locals {
   master_password      = var.create_db_instance && var.create_random_password ? random_password.master_password[0].result : var.password
-  db_subnet_group_name = var.replicate_source_db != null ? null : coalesce(var.db_subnet_group_name, module.db_subnet_group.db_subnet_group_id, var.identifier)
+  db_subnet_group_name = var.cross_region_replica != true && var.replicate_source_db != null ? null : coalesce(var.db_subnet_group_name, module.db_subnet_group.db_subnet_group_id, var.identifier)
```
antonbabenko, line 3: @bryantbiggs `local.db_subnet_group_name` does not seem to be used in module `db_subnet_group` on line 24. Is it correct?

antonbabenko (commented):
I'd like to ask what @bryantbiggs thinks about the name for `cross_region_replica` variable and this PR?

main.tf
bryantbiggs: correct - we only create the subnet group when `create_db_subnet_group = true` and the name is either `db_subnet_group_name` or `identifier`. 

then, based on the logic on line 3 - we provide the subnet group name to the instance itself https://github.com/terraform-aws-modules/terraform-aws-rds/blob/e2636f94f5e7d90a0b9480bf73fe0268ffe372cf/main.tf#L89

riccardo-salamanna: Yeah the issue is that if want to create a read only replica, db_subnet_group_name defaults to null, which is necessary for same region replicas but it is not correct for cross-region (different vpc, different subnets etc etc). I proposed the use of a different variable so that we can override that behaviour, and at the same time allow the use of an existing subnet group or the creation of a new one with the modukle, preserving the functionality of the module itself as is. Thank you :)

antonbabenko: @bryantbiggs WDYT? yes/no? :)

daniel-huesca (approved):

cristian041 (approved):

antonbabenko:
¡Hola todos! Great to see so many approvals from Barcelona 🇪🇸 but let's hear if Bryant approves this, too :)

main.tf
bryantbiggs: I think this is ok - maybe we could change `var.cross_region_replica != true && ...` to `!var.cross_region_replica && ...`

we could also update the replica examples to add an instance in another region to ensure this works (and continues to work as intended with any future changes) https://github.com/terraform-aws-modules/terraform-aws-rds/blob/e2636f94f5e7d90a0b9480bf73fe0268ffe372cf/examples/replica-postgres/main.tf#L112

riccardo-salamanna: I'm willing to do that for you if that is ok :)

variables.tf
```diff
@@ -33,6 +33,12 @@ variable "replicate_source_db" {
   default     = null
 }
 
+variable "cross_region_replica" {
+  description = "Specifies if the replica should be cross region"
+  type        = string
```
antonbabenko, line 38:
```suggestion
  type        = bool
```

Right?

variables.tf
```diff
@@ -33,6 +33,12 @@ variable "replicate_source_db" {
   default     = null
 }
 
+variable "cross_region_replica" {
+  description = "Specifies if the replica should be cross region"
```
antonbabenko, line 37:
```suggestion
  description = "Specifies if the replica should be cross region. It allows the use of a subnet group in a region different than the master instance"
```

antonbabenko (approved):
LGTM

variables.tf
riccardo-salamanna: right indeed 🤦

antonbabenko:
Vamos! :)

[v3.3.0](https://github.com/terraform-aws-modules/terraform-aws-rds/releases/v3.3.0) has been just released.

riccardo-salamanna:
Thanks Anton for fixing it all!!!!! a pleasure. have a great day

PR #355, opened 2021-09-20
Add db_instance_host output

FieryCod:
## Description
The following change adds the additional `db_instance_host`. The proposed parameter is useful when obtaining the db auth token from the codepipeline in which the `db_instance_host` is injected:

https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/UsingWithRDS.IAMDBAuth.Connecting.AWSCLI.PostgreSQL.html

## Motivation and Context

Property `db_instance_endpoint` contains the full connection URL, but for `aws rds generate-db-auth-token` the port should specified separately. The following change adds a new property that is the `db_instance_endpoint` - `db_instance_port`. 

## Breaking Changes

No breaking changes.

## How Has This Been Tested?
Not required.

### DIFF
--- a/outputs.tf
+++ b/outputs.tf
@@ -28,6 +28,11 @@ output "db_instance_endpoint" {
   value       = module.db_instance.db_instance_endpoint
 }
 
+output "db_instance_host" {
+  description = "The database host"
+  value       = replace(module.db_instance.db_instance_endpoint, "/:.*/", "")
+}
+
 output "db_instance_hosted_zone_id" {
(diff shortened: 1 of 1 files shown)


gthieleb:
There already exists the output property `db_instance_address` that is the FQDN to the database *without* the port number.

antonbabenko:
`db_instance_address` is here https://github.com/terraform-aws-modules/terraform-aws-rds/blob/master/outputs.tf#L11-L14

PR #358, opened 2021-10-18
chore: update CI workflow to utilize new composite actions and format with terraform-docs v0.16.0

bryantbiggs:
## Description
- update CI workflow to utilize new composite actions with terraform-docs v0.16.0 (default version for v1.1.0 of `pre-commit` composite action)

~:warning: A GitHub personal access token will need to be created and either added to just this repo for now, or added as an organization secret (defined in one place and available across all repos in the org). You can see more here https://github.com/clowdhaus/terraform-composite-actions/tree/main/commit#github-token-permissions - this is to enable the automatic formatting of PRs. To be transparent - I don't know how this will fair on PRs that do not check the `Allow edits and access to secrets by maintainers` checkbox. The workflow is designed to allow for failures without failing the whole CI so it should be ok~

unfortunately I don't see a secure way around the authentication flow on forked copies. either the `terraform-aws-modules` organization has to extend access to forks to use the token created, or users must create their own token and upload it to their forked projects. The first option won't work because we don't want to expose access to a writable token, and the latter seems less likely to happen by users unfortunately. Therefore, I have removed this functionality from the PR

## Motivation and Context
- consolidate CI workflow logic

## Breaking Changes
- No

## How Has This Been Tested?
- See test PR https://github.com/clowdhaus/terraform-release-test/pull/8

### DIFF
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -6,6 +6,9 @@ on:
       - main
       - master
 
+env:
+  TERRAFORM_DOCS_VERSION: v0.16.0
+
 jobs:
   collectInputs:
     name: Collect workflow inputs
@@ -43,6 +46,7 @@ jobs:
         uses: clowdhaus/terraform-composite-actions/pre-commit@v1.3.0
         with:
           terraform-version: ${{ steps.minMax.outputs.minVersion }}
+          terraform-docs-version: ${{ env.TERRAFORM_DOCS_VERSION }}
           args: 'terraform_validate --color=always --show-diff-on-failure --files ${{ matrix.directory }}/*'
 
       - name: Pre-commit Terraform ${{ steps.minMax.outputs.minVersion }}
@@ -51,6 +55,7 @@ jobs:
         uses: clowdhaus/terraform-composite-actions/pre-commit@v1.3.0
         with:
           terraform-version: ${{ steps.minMax.outputs.minVersion }}
+          terraform-docs-version: ${{ env.TERRAFORM_DOCS_VERSION }}
           args: 'terraform_validate --color=always --show-diff-on-failure --files $(ls *.tf)'
 
   preCommitMaxVersion:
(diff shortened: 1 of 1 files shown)


bryantbiggs:
The version of `terraform-docs` is specified as an input variable [here](https://github.com/clowdhaus/terraform-composite-actions/blob/eb963bdb7f8b1c64e29b6f4b08c7e5eab1be4bd2/pre-commit/action.yml#L10) where the current default is v0.16.0 which is the latest. 

you can also update the action to specify which version should be used as well, something like:
```yml
...
	- name: Pre-commit Terraform ${{ steps.minMax.outputs.maxVersion }}
        uses: clowdhaus/terraform-composite-actions/pre-commit@v1.3.0
        with:
          terraform-version: ${{ steps.minMax.outputs.maxVersion }}
          terraform-docs-version: v0.12.1 # or whatever version is desired
```

however, I will say - running `terraform-docs` through pre-commit has been a horrible headache. pre-commit is designed to work of changes that are staged for the next commit, and so running it in this type of workflow can be problematic as opposed to using the tool directly.

for example:
scenario 1 (ideal): user creates changes, does `git commit -m "my changes"` and pre-commit will check the staged changes and format/update docs accordingly

scenario 2: user creates changes and bypasses commit hooks using the `-n` flag like `git commit -m "my changes"`. if you then try to run manually with `pre-commit run -a`, pre-commit may or may not detect the changes and update (I have not been able to identify a pattern). whereas using the terraform-docs tool directly, it picks up the changes everytime regardless, same with terraform fmt

to be clear, this isn't a limitation of the workflow or actions shown here - they are merely just calling pre-commit as requested and therefore pre-commit is the uknown here that is causing these weird issues

bryantbiggs:
it currently is v0.16.0 because that is what is what is specified at the time of the v1.3.0 release of the action. if a new version comes out, nothing will happen here

If there is a new release of `terraform-docs` we have two options, either route will require a PR to change the repository to use the different version:
a. The default version is updated [here](https://github.com/clowdhaus/terraform-composite-actions/blob/eb963bdb7f8b1c64e29b6f4b08c7e5eab1be4bd2/pre-commit/action.yml#L10), a new release of the actions is created, and a PR is opened against a module repo to use the newly released version which also contains the newly released version of `terraform-docs`
b. We specify the version in the module repo using the `terraform-docs-version: v#.#.#`, when a new version of `terraform-docs` is released we open a PR to update the `v#.#.#` version

either route, changes are tied to a release of some sorts; there is no "floating release" that can come into play unknowningly. I'll update to add the version here as well

PR #359, opened 2021-10-19
fix: Add full subnet group name when prefix is enabled

david92rl:
## Description
Currently, it's not possible to specify a prefix for the subnet group because `coalesce(var.db_subnet_group_name, module.db_subnet_group.db_subnet_group_id, var.identifier)` will return a simple name even if `db_subnet_group_use_name_prefix` is set to true.
This makes sure we always use the right sunet group id.

## Breaking Changes
This is backwards compatible

## How Has This Been Tested?

Used the `examples/complete-postgres` example to make sure the db created uses the right subnet group id


main.tf
```diff
@@ -1,6 +1,6 @@
 locals {
   master_password      = var.create_db_instance && var.create_random_password ? random_password.master_password[0].result : var.password
-  db_subnet_group_name = !var.cross_region_replica && var.replicate_source_db != null ? null : coalesce(var.db_subnet_group_name, module.db_subnet_group.db_subnet_group_id, var.identifier)
+  db_subnet_group_name = ! var.cross_region_replica && var.replicate_source_db != null ? null : var.create_db_subnet_group ? module.db_subnet_group.db_subnet_group_id : coalesce(var.db_subnet_group_name, var.identifier)
```
bryantbiggs, line 3: this can actually be more succinctly fixed with:

```suggestion
  db_subnet_group_name = !var.cross_region_replica && var.replicate_source_db != null ? null : try(module.db_subnet_group.db_subnet_group_id, var.db_subnet_group_name)
```

what this does is:
1. Try to get the computed subnet group ID, if that fails
2. Use the `var.db_subnet_group_name`

I can't see a scenario where we don't can get past both the created or provided subnet group names so we can drop the `var.identifier` here

elovelan:
@david92rl, I'm going to give this a go 😃

antonbabenko:
This issue has been resolved in version 4.0.0 :tada:

PR #388, opened 2022-02-23
fix: Correct random password lifecycle by eagerly creating random password value

### LINKED ISSUE #386: Variable interdependecy casues terraform plan to fail in 4.1.2
4.1.2 introduced a situation whereby running terraform plan is not possible because the local variable ` create_random_password`
```
create_random_password = var.create_db_instance && var.create_random_password && var.replicate_source_db == null
```
is evaluated based on the dynamic `var.replicate_source_db` whose value cannot be determined until after terraform apply. 

`replicate_source_db` is usually referenced by another db instance that is typically created in the same terraform job, example:  
`replicate_source_db = module.rds.db_instance_id`

Terraform cannot tell whether `module.rds.db_instance_id` will be null or not thus it fails to render a plan.

bryantbiggs:
## Description
- Correct random password lifecycle by eagerly creating random password value. Instead of linking the random password resource generation to whether or not the instance is a replica in the root, just let the random password eagerly create a value; whether or not the password value generated is used is handled by the `db_instance` sub-module regardless https://github.com/terraform-aws-modules/terraform-aws-rds/blob/master/modules/db_instance/main.tf#L8

## Motivation and Context
- Fix lifecycle dependency issue causing errors when trying to use replicas
- Closes #386 

## Breaking Changes
- No

## How Has This Been Tested?

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -1,6 +1,6 @@
 locals {
-  create_random_password = var.create_db_instance && var.create_random_password && var.replicate_source_db == null
-  password               = try(random_password.master_password[0].result, var.password)
+  create_random_password = var.create_db_instance && var.create_random_password
+  password               = local.create_random_password ? random_password.master_password[0].result : var.password
 
   db_subnet_group_name    = var.create_db_subnet_group ? module.db_subnet_group.db_subnet_group_id : var.db_subnet_group_name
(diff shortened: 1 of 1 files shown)


antonbabenko:
This PR is included in version 4.1.3 :tada:

PR #421, opened 2022-07-22
Feature/added random password value

ravindrasinghh:
## Description
I have removed to put the manual password, instead of that we can create random password

## Motivation and Context
To enhance more security

## Breaking Changes
It will not break any changes.

## How Has This Been Tested?
yes


bryantbiggs:
we already have support for this https://github.com/terraform-aws-modules/terraform-aws-rds/blob/master/main.tf#L16

This change would break users configurations

ravindrasinghh:
okay,Got it 
Thanks @bryantbiggs

PR #422, opened 2022-07-22
Feature/store password in ssm

ravindrasinghh:
## Description
Store DB admin password in the AWS SSM parameter

## Motivation and Context
We can store the password in SSM paramter instead of terraform TF state.

## Breaking Changes
There will be no breaking changes. and no user will impact by adding this code.

## How Has This Been Tested?
yes

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -24,6 +24,7 @@ resource "aws_ssm_parameter" "master_db_password" {
   name   = "/rds/var.identifier/password"
   value  = var.create_random_password == true ? random_password.master_password[0].result : var.password
   type   = "SecureString"
+  key_id = var.ssm_kms_key
 }
     
 module "db_subnet_group" {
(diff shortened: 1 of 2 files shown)


bryantbiggs:
Thanks for the PR but I don't think this is something thats necessary here - users are able to take the password output and store it in SSM, SecretsManager, etc. and that can be implemented outside of the module

ravindrasinghh:
>users are able to take the password output and store it in SSM, SecretsManager, etc.
That's why I added the above code. So they can store the password in SSM and in statefile too.
This is an enhanced feature. So people can take a reference of the above code to store the admin password in SSM.
Adding the above code will not make any breaking changes to the existing infrastructure.
We can also make it more dynamic as a feature like:

resource "aws_ssm_parameter" "master_db_password" {
   name   = "/rds/var.identifier/password"
   value  = var.create_random_password == true && var.store_password_ssm == true ? random_password.master_password[0].result : var.passwordrandom_password.master_password[0].result : var.password
   type   = "SecureString"
   key_id = var.ssm_kms_key
 }
 
 variable "store_password_ssm" {
 type    = bool
 default = false
 }
 
 Please suggest to me if anything needs to change.

bryantbiggs:
No changes required, I guess what I am saying is that I don't see this as a feature we want to add here and instead leave it up to users as to how they want to manage their password

ravindrasinghh:
okak

Issue #408, opened 2022-06-07
Can you provide examples of Multi AZ DB cluster?


ianeinser:
## Is your request related to a new offering from AWS?
Is this functionality available in the [AWS provider for Terraform](https://registry.terraform.io/providers/hashicorp/aws/latest/docs)? See [CHANGELOG.md](https://github.com/hashicorp/terraform-provider-aws/blob/main/CHANGELOG.md), too.
  - No 🛑: please wait to file a request until the functionality is avaialble in the AWS provider
  - Yes ✅: please list the AWS provider version which introduced this functionality

antonbabenko:
You have to specify [`multi_az = true`](https://github.com/terraform-aws-modules/terraform-aws-rds#input_multi_az) as a module argument, and AWS RDS service will do the rest.

rickychew77:
hi @antonbabenko, currently there were 3 deployment options based on [RDS docs here](aws.amazon.com/rds/features/multi-az/). Also can be seen in AWS console while provisioning in UI.

1. Single-AZ
2. Multi-AZ with one standby (the standby just act as failover, does not serve read traffic)
3. Multi-AZ with two readable standby 

Using `multi_az = true` seems like deploying #2 options above. Does RDS terraform provider support option #3 above? 

Saying so because I've just tried using `multi_az = true` and it was able to create with instance class `db.t3.micro` and #3 option only supports m and r classes (as seen in console UI).

agomezvidalee:
> hi @antonbabenko, currently there were 3 deployment options based on [RDS docs here](https://aws.amazon.com/rds/features/multi-az/) (see comparison table in the link). Also can be seen in AWS console while provisioning in UI.
> 
> 1. Single-AZ
> 2. Multi-AZ with one standby (the standby just act as failover, does not serve read traffic)
> 3. Multi-AZ with two readable standby
> 
> Using `multi_az = true` seems like deploying option 2 above. Does RDS terraform provider support option 3 above?
> 
> Saying so because I've just tried using `multi_az = true` and it was able to create with instance class `db.t3.micro` and option 3 only supports m and r classes (as seen in console UI).

which engine do you deploy?

rickychew77:
I am deploying to PostgreSQL

agomezvidalee:
> I am deploying to PostgreSQL @agomezvidalee

[...]

PR #424, opened 2022-08-01
feat: Multi AZ RDS Cluster Support

magreenbaum:
## Description
Support for multi-az rds clusters (mysql and postgres). 
Added in aws provider version 4.6: https://github.com/hashicorp/terraform-provider-aws/pull/23684

Additional comments: 
There are a lot of limitations on non-aurora rds clusters at the moment. Only attributes that are supported for non-aurora rds clusters are included in this PR. Aurora RDS cluster module referenced for this feature: https://github.com/terraform-aws-modules/terraform-aws-rds-aurora

Refs: 
https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/create-multi-az-db-cluster.html
https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/multi-az-db-clusters-concepts.html

## Motivation and Context
https://github.com/terraform-aws-modules/terraform-aws-rds/issues/408

## Breaking Changes
No.

## How Has This Been Tested?

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,9 +7,12 @@ Terraform module which creates RDS resources on AWS.
 Root module calls these modules which can also be used separately to create independent resources:
 
 - [db_instance](https://github.com/terraform-aws-modules/terraform-aws-rds/tree/master/modules/db_instance) - creates RDS DB instance
+- [db_instance_automated_backups_replication](https://github.com/terraform-aws-modules/terraform-aws-rds/tree/master/modules/db_instance_automated_backups_replication) - creates RDS DB automated backup replication
 - [db_subnet_group](https://github.com/terraform-aws-modules/terraform-aws-rds/tree/master/modules/db_subnet_group) - creates RDS DB subnet group
 - [db_parameter_group](https://github.com/terraform-aws-modules/terraform-aws-rds/tree/master/modules/db_parameter_group) - creates RDS DB parameter group
 - [db_option_group](https://github.com/terraform-aws-modules/terraform-aws-rds/tree/master/modules/db_option_group) - creates RDS DB option group
+- [rds_cluster](https://github.com/terraform-aws-modules/terraform-aws-rds/tree/master/modules/rds_cluster) - creates Multi-AZ AZ cluster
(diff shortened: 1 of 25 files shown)


bryantbiggs (changes requested):
I'm not sure I understand whats going on in this PR 
- `aws_rds_cluster` is an Aurora cluster and we have a separate module for that
- I don't think we need to add the KMS key module and multi-region to each example, nor do I understand how it relates to multi-AZ cluster support

magreenbaum:
> 

The terraform-aws-provider recently added support for non-aurora engines (mysql & postgres). Ref: https://github.com/hashicorp/terraform-provider-aws/pull/23684
https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/create-multi-az-db-cluster.html
https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/multi-az-db-clusters-concepts.html
I think using the aurora module would work (some attributes would need to be added) but the naming is a bit confusing. If someone is looking to create a non-aurora mysql rds multi-az cluster for example, looking here might make more sense.

KMS key module was added to the cross-region-replica-postgres example due to an error:
Screen Shot 2022-07-30 at 10 53 55 PM
A KMS key in the destination region must be specified when the source DB is encrypted. 
I can remove this from the PR and make it a separate fix if that is preferable. 

I added examples for both multi-az cluster supported engine types but I can remove one since we don't need both.

bryantbiggs:
hmm, I'm not sure how best to proceed. It seems silly to replicate [aws_rds_cluster](https://github.com/terraform-aws-modules/terraform-aws-rds-aurora/blob/master/main.tf#L47) here just to support multi-AZ RDS, plus that now opens the door for other cluster resources which just exacerbates the situation. 

I'm more inclined to just show an example of how to use the Aurora module to create a multi-AZ RDS cluster (non-Aurora). @antonbabenko what are your thoughts?

antonbabenko:
I agree with @bryantbiggs. There is already a separate module and if necessary (looks like it is a case) let's update just an example in this one.

If something is missing in another RDS module we have, please open a small PR there.

magreenbaum:
Understood, that makes sense, thanks for the feedback @bryantbiggs @antonbabenko.

PR #431, opened 2022-09-10
feat: Add support for `network_type`

krewenki:
## Description
Add support for the `network_type` argument.

## Motivation and Context
I've got RDS clusters in production that need to be deployed in dual-stack mode, and this module didn't support that configuration yet.  Instead of maintaining a private fork, I will submit a patch.

## Breaking Changes
The required AWS provider is bumped from 4.6 to 4.28.  This version is where the network_type argument became available.

## How Has This Been Tested?

### DIFF
--- a/modules/db_instance/variables.tf
+++ b/modules/db_instance/variables.tf
@@ -349,7 +349,7 @@ variable "restore_to_point_in_time" {
 variable "network_type" {
   description = "The type of network stack"
   type        = string
-  default     = "IPV4"
+  default     = null
 }
 
 ################################################################################
--- a/modules/db_instance/variables.tf
+++ b/modules/db_instance/variables.tf
@@ -346,6 +346,12 @@ variable "restore_to_point_in_time" {
   default     = null
 }
 
+variable "network_type" {
+  description = "The type of network stack"
+  type        = string
+  default     = "IPV4"
+}
+
 ################################################################################
 # CloudWatch Log Group
 ################################################################################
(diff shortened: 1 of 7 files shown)


modules/db_instance/variables.tf
```diff
@@ -346,6 +346,12 @@ variable "restore_to_point_in_time" {
   default     = null
 }
 
+variable "network_type" {
+  description = "The type of network stack"
+  type        = string
+  default     = "IPV4"
```
bryantbiggs, line 352:
```suggestion
  default     = null
```

variables.tf
```diff
@@ -478,6 +478,12 @@ variable "random_password_length" {
   default     = 16
 }
 
+variable "network_type" {
+  description = "The type of network stack to use"
+  type        = string
+  default     = "IPV4"
```
bryantbiggs, line 484:
```suggestion
  default     = null
```

bryantbiggs (changes requested):

modules/db_instance/variables.tf
krewenki: This has been changed. I'll leave it up to your review to mark these conversations as resolved.

bryantbiggs (approved):
thank you!

antonbabenko:
This PR is included in version 5.1.0 :tada:

chenrui333:
QQ, does changing `network_type` would require DB bouncing or re-creation?

antonbabenko:
@chenrui333 I assume that if you are not changing the value itself (but rather importing existing value), it should not prompt you to recreate anything. I have not tried this myself.

PR #456, opened 2022-12-14
fix: Change RDS variable `iops` default value from `0` to `null`

etiennechabert:
## Description
The RDS variable `iops` is currently having a default of `0`. 
In this PR, I propose that the default value become `null` and so to avoid side effects on the AWS Provider side.

## Motivation and Context
The current default value is causing side effects for people currently trying to adopt `gp3` (see: https://github.com/hashicorp/terraform-provider-aws/issues/28271)

## Breaking Changes
No idea, could be, but I have the feeling not.

## How Has This Been Tested?
Not tested yet, just proposing the idea


modules/db_instance/variables.tf
```diff
@@ -180,7 +180,7 @@ variable "multi_az" {
 variable "iops" {
   description = "The amount of provisioned IOPS. Setting this implies a storage_type of 'io1'"
   type        = number
-  default     = 0
+  default     = null
```
etiennechabert, line 183: This is a very candide proposal, I have no idea if there is a good reason or not for the default value currently been `0`.

The only thing I know: this is currently causing an incompatibility with the default behavior of the AWS Provider in his version 4.46.0.

Someone, not using `terraform-aws-rds` module, can directly transition from `gp2` to `gp3` (see: https://github.com/hashicorp/terraform-provider-aws/issues/28271#issuecomment-1348194124)

But using `terraform-aws-rds`, you are forced to override the default value of `iops` from `0`, to `null`:
https://github.com/hashicorp/terraform-provider-aws/issues/28271#issuecomment-1349076641

bryantbiggs: its optional so `null` should be the appropriate default value

bryantbiggs:
you'll need to run `pre-commit run terraform_docs -a` to update the docs accordingly

bryantbiggs (approved):
thank you!

etiennechabert:
Thanks for your help putting it together 🤝

antonbabenko:
This PR is included in version 5.2.1 :tada:

lucasvuotto:
Hi! I don't know if this is the best place to comment this (please point me to the right one if it's the case), but this PR kinda misses one case: the change was only applied to `db-instance` submodule. If instead one is using the root module, the default for `iops` is still `0` instead of `null`, and this `0` gets propagated to `db-instance`.

etiennechabert:
Thanks for pointing this out @lucasvuotto, will replicate my recent changes in the main module, starting with:
https://github.com/terraform-aws-modules/terraform-aws-rds/pull/462

Issue #495, opened 2023-05-31
add identifier to outputs


shilikok:
## Is your request related to a new offering from AWS?
Is this functionality available in the [AWS provider for Terraform](https://registry.terraform.io/providers/hashicorp/aws/latest/docs)? See [CHANGELOG.md](https://github.com/hashicorp/terraform-provider-aws/blob/main/CHANGELOG.md), too.
  - Yes ✅: please list the AWS provider version which introduced this functionality
aws provider: 5.0.0

## Is your request related to a problem? Please describe.
id is no longer the AWS database identifier - id is now the dbi-resource-id. Refer to identifier instead of id to use the database's identifier

## Describe the solution you'd like.
add to module outputs db identifier value

bryantbiggs:
while the attribute may have always been there, the value for `id` has changed between < v4.x & > v5.x 

so while simply adding the `identifier` output seems harmless, I don't think that is providing value without addressing this underlying change

daroga0002:
issue is that database name is not returned on v5 provider by any parameter. After adding this value it module will become working correctly with v5 provider

daroga0002:
@shilikok  FYI

I am currently adressing this by making workaround by extracting from instance ARN last element which represent identifier of database

```hcl
module "database_replica_rds" {
  count   = local.create_database_replica_rds ? 1 : 0
  source  = "terraform-aws-modules/rds/aws"
  version = "5.9.0"

  # Source database. For cross-region use db_instance_arn
  replicate_source_db        = element(split(":", module.database_rds[0].db_instance_arn), length(split(":", module.database_rds[0].db_instance_arn))-1)
  .....
}
```

AyodejiO:
I just had the same issue where I had to manually use the `local.identifier` variable used to create the master DB. You should prolly use that to remove the complex code you have @daroga0002

daroga0002:
> I just had the same issue where I had to assign the `replicate_source_db` to the `local.identifier` variable used to create the master DB. You should prolly use that to remove the complex code you have @daroga0002

It will be not working when you use [instance_use_identifier_prefix](https://github.com/terraform-aws-modules/terraform-aws-rds#input_instance_use_identifier_prefix) variable.

What is more it will require additional manual dependency establish between main and replica module (when you use outputs from one module dependency is done automaticlly)

daroga0002:
@bryantbiggs @antonbabenko  
[...]

PR #499, opened 2023-06-20
fix: Update to follow change in db_instance id

jwoytek:
## Description
A breaking change in the AWS provider changed the output of the upstream `id` attribute. Update the `id` module output to return the RDS DB instance identifier. This is an alternate solution to the one proposed in #495.

## Motivation and Context
`db_instance` was updated by https://github.com/hashicorp/terraform-provider-aws/pull/31232 such that the `id` attribute now tracks the resource ID, and the identifier is now returned in a new attribute named `identifier`.

The resource ID is already available in `instance_resource_id`. This fix allows existing users to continue using this module with no changes and preserves the original function of the `id` attribute in this module.

Fixes the base issue reported in #495.

## Breaking Changes
This should not represent a breaking change, as the referenced `identifier` attribute has
existed in `db_instance` since 2.33, which is well earlier than the current requirements for
this module. 

## How Has This Been Tested?
- [X] I have tested and validated these changes using one or more of the provided `examples/*` projects

### DIFF
--- a/modules/db_instance/outputs.tf
+++ b/modules/db_instance/outputs.tf
@@ -50,7 +50,7 @@ output "db_instance_hosted_zone_id" {
 
 output "db_instance_id" {
   description = "The RDS instance ID"
-  value       = try(aws_db_instance.this[0].id, "")
+  value       = try(aws_db_instance.this[0].identifier, "")
 }
 
 output "db_instance_resource_id" {


antonbabenko:
This issue has been resolved in version 6.0.0 :tada:

PR #516, opened 2023-10-10
feat: Add possibility to overwrite Name tag

widgetpl:
## Description
Sometimes you need to import existing resources which have different Name and Tag with Name key.
Especially when you are importing `db_subnet_group` current state forces to recreate this resource which should be in different VPC to make it work.

## Motivation and Context
Be able to overwrite Name tag for below resources:
- `aws_db_parameter_group`
- `aws_db_subnet_group`
- `aws_db_option_group`

Due to behaviour of [`merge()`](https://developer.hashicorp.com/terraform/language/functions/merge) function we are not able to overwrite `Name` tag. Changing order will allow this:
>If more than one given map or object defines the same key or attribute, then the one that is later in the argument sequence takes precedence.

## Breaking Changes
None

## How Has This Been Tested?

### DIFF
--- a/modules/db_option_group/main.tf
+++ b/modules/db_option_group/main.tf
@@ -34,10 +34,10 @@ resource "aws_db_option_group" "this" {
   }
 
   tags = merge(
-    var.tags,
     {
       "Name" = var.name
     },
+    var.tags,
   )
 
   timeouts {
--- a/modules/db_parameter_group/main.tf
+++ b/modules/db_parameter_group/main.tf
@@ -23,10 +23,10 @@ resource "aws_db_parameter_group" "this" {
   }
 
   tags = merge(
-    var.tags,
     {
       "Name" = var.name
     },
+    var.tags,
   )
 
   lifecycle {
--- a/modules/db_subnet_group/main.tf
+++ b/modules/db_subnet_group/main.tf
@@ -14,9 +14,9 @@ resource "aws_db_subnet_group" "this" {
   subnet_ids  = var.subnet_ids
 
   tags = merge(
-    var.tags,
     {
       "Name" = var.name
     },
+    var.tags,
   )
 }


widgetpl:
@antonbabenko can you take a look?

bryantbiggs:
I don't know that this is something we want to change - it would be odd to have a resource named `foo` but the tag on it is `{Name = bar}`. if the issue is that the subnet group that was created outside of Terraform does not conform with this module, I would suggest either leaving it outside of this module or let this module create the subnet group with the desired settings

widgetpl:
@bryantbiggs I agree with you, but this change isn't changing anything unless you consciously define `Name` tag.
The case for me is that I am not able to recreate/create `aws_db_subnet_group` as this forces to have new subnet group in different VPC. So I wanted to make it optional to overwrite `Name` tag.

widgetpl:
@bryantbiggs I see that there are different ways to define tags for module resources, like for subnets in VPC module we have:
```
tags = merge(
    {
      Name = try(
        var.public_subnet_names[count.index],
        format("${var.name}-${var.public_subnet_suffix}-%s", element(var.azs, count.index))
      )
    },
    var.tags,
    var.public_subnet_tags,
    lookup(var.public_subnet_tags_per_az, element(var.azs, count.index), {})
  )
```
so in this case you are able to overwrite `Name` tag. As I said this change is backward compatible and adds only simple feature to overwrite `Name` tag when you need it (optional).

widgetpl:
@bryantbiggs can we have this change? It doesn't change anything as I said until you want to have different value for `Name` tag. It really just can help people to import this resources especially `db_subnet_group` which you can't update just like that for RDS.

bryantbiggs:
unfortunately this has the potential to affect users, and it doesn't align with how the module was designed so we will not be accepting this change

Issue #530, opened 2024-01-10
The psql password passed to the module is overridden by the one aws creates in Secret Manager


lorenzophys:
## Description

When creating a PostgreSQL RDS instance using the `terraform-aws-rds` module, I encountered an unexpected behavior where AWS auto-generates a password in the Secret Manager, even though I explicitly provided a username and password. This occurs despite the documentation stating that the `manage_master_user_password` input is false by default.
If I explicitly set `manage_master_user_password = false` then my password is actually used for the master user.

## Versions
- Module version [Required]:

- Terraform version:
```
Terraform v1.5.7
on darwin_arm64
```
- Provider version(s):
```
Terraform v1.5.7
on darwin_arm64
+ provider registry.terraform.io/carlpett/sops v1.0.0
+ provider registry.terraform.io/hashicorp/aws v5.31.0
+ provider registry.terraform.io/hashicorp/random v3.6.0
```
## Reproduction Code [Required]

```hcl
module "db" {
  source = "terraform-aws-modules/rds/aws"

  identifier = local.id

  engine                    = "postgres"
  engine_version            = "12.14"
  major_engine_version      = "12"
  family                    = "postgres12" # DB parameter group
  create_db_parameter_group = false

  # DB option group
  instance_class    = "db.t3.small"
  allocated_storage = 20

  username                    = "postgres"
  password                    = data.sops_file.db_credentials.data["secrets.db_password"]
  #manage_master_user_password = false
  port                        = 5432

  multi_az               = false
  vpc_security_group_ids = [module.security_group.security_group_id]

  create_db_subnet_group      = true
  db_subnet_group_name        = "${local.short_name}-${data.aws_vpc.eks.id}"
  db_subnet_group_description = "Database subnet group accociated with ${local.environment} eks cluster for ${local.short_name}"
  subnet_ids                  = [for subnet in data.aws_subnets.eks.ids : subnet if !(contains(data.aws_subnets.eks_public.ids, subnet))]

  iam_database_authentication_enabled = false
  publicly_accessible                 = false

  auto_minor_version_upgrade = true

  maintenance_window = "Mon:00:00-Mon:03:00"
  backup_window      = "03:00-06:00"

  backup_retention_period = 0
  skip_final_snapshot     = true
  deletion_protection     = false

  performance_insights_enabled = false

  tags = local.tags
}
```

## Steps to reproduce the behavior:

[...]

PR #531, opened 2024-01-10
fix: Align the default value of `manage_master_user_password` in the instance sub-module with the root module

lorenzophys:
## Description
Fixed the default value of `manage_master_user_password` in the `db_instance` submodule to avoid confusion. See #530


bryantbiggs (approved):
thank you!

For reference - https://github.com/terraform-aws-modules/terraform-aws-rds/issues/530#issuecomment-1884898941

PR #534, opened 2024-01-15
fix: Allow managing `allocated_storage` for replicas

relu:
## Description

A regression was introduced via c66cd738 where allocated_storage can only be set on primary instances. However, this should not be a limitation because RDS does allow managing replica storage independently from the primary.


antonbabenko:
This PR is included in version 6.5.1 :tada:

PR #544, opened 2024-03-18
fix: Separate `db_instance_tags` from merged `tags` sub-module input

winthropgutmann-ef:
## Description
 This PR splits the two on input so that the `var.db_instance_tags` are only applied to the rds_db_instance, preserving the functionality that if `var.tags` is defined, it will be merged with `var.db_instance_tags`

## Motivation and Context
This change is required because it incorrectly maps `var.db_instance_tags` to all resources associated with the module. This functionality is reserved for the `var.tags` input. `var.db_instance_profile` described as "Additional tags for the DB instance", only these tags should be associated with the `rds_db_instance`

## Breaking Changes
None - backwards compatible since `var.db_instance_tags` and `var.tags` are still merged on `this.db_instance` when present.

## How Has This Been Tested?

I need to test these changes, we have an existing project that this might be applicable in. 

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,6 @@
 repos:
   - repo: https://github.com/antonbabenko/pre-commit-terraform
-    rev: v1.88.0
+    rev: v1.88.2
     hooks:
       - id: terraform_fmt
       - id: terraform_docs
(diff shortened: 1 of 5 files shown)


antonbabenko:
This PR is included in version 6.5.3 :tada:

Issue #510, opened 2023-09-07
Show error if password is provided and manage_master_user_password is true


TeemuKoivisto:
## Description

Hey, so as I was attempting to connect to my RDS instance I got stuck trying to establish the SSL connection. While that was already quite annoying it really boggled my mind why I still couldn't connect to the instance. Turns out, while I was providing a DB password with `password = var.db_password` this was never used as `manage_master_user_password` is set to true by default.

While it might be a best practise to do this by default, I'd really want an error shown about the discarded password. It's very unexpected behavior and I'd even categorize it as a bug.

- [x] ✋ I have searched the open/closed issues and my issue is not listed.

## Versions

- Module version [Required]: "6.1.1"

- Terraform version: v1.5.6

- Provider version(s): aws 5.15.0

PR #565, opened 2024-09-30
feat: Automatically disable `manage_master_user_password` when `password` is provided

SoulMind01:
## Description

This pull request adds functionality to automatically set `manage_master_user_password` to `false` when the user provides a manual `password`. This ensures that the manually provided password is respected and not overridden by Terraform when `manage_master_user_password` is set to `true` by default.

## Motivation and Context

This change resolves an issue where users provide a password, but it is ignored because `manage_master_user_password` is set to `true` by default. By automatically setting `manage_master_user_password` to `false` when a password is provided, the module behaves more intuitively, using the password provided by the user.

This change addresses an issue discussed in [#510](https://github.com/terraform-aws-modules/terraform-aws-rds/issues/510), where the unexpected behavior caused confusion.

## Breaking Changes

This change is backward-compatible. If no password is provided, the module will behave as it currently does, with `manage_master_user_password` defaulting to `true`. This ensures that the new behavior only affects users who explicitly provide a password.

## How Has This Been Tested?
+ I have validated the changes is working correctly by testing in `examples/complete-postgres`. After providing a value to `password`, `manage_master_user_password` is automatically set to false.


SoulMind01:
Apologies for making changes directly on the master branch. I'll create a new branch and submit a new pull request shortly.

PR #574, opened 2024-10-30
adding dynamic lifecycle rule to ignore username/password changes if …

JamesDavidson13:
…var.password is declared

## Description
I added a dynamic lifecycle rule that will trigger when var.password is declared, resulting in ignoring changes for username and password.

## Motivation and Context
I need to enable ignore_changes for the username and password of my DB, so i can merge my code without sensitive values.

## Breaking Changes
N/A

## How Has This Been Tested?
- [ x] I have updated at least one of the `examples/*` to demonstrate and validate my change(s)
- [ x] I have tested and validated these changes using one or more of the provided `examples/*` projects

- [x ] I have executed `pre-commit run -a` on my pull request

### DIFF
--- a/modules/db_instance/main.tf
+++ b/modules/db_instance/main.tf
@@ -145,6 +145,17 @@ resource "aws_db_instance" "this" {
     update = lookup(var.timeouts, "update", null)
   }
 
+  dynamic "lifecycle" {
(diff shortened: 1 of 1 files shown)


bryantbiggs:
unfortunately we cannot support this because there are folks who want to continue setting the username and/or password through Terraform

PR #589, opened 2025-03-05
feat: Add support for write-only password

jvanbrunschot:
## Description
Add support for write_only passwords. There seems to be a bug in terraform 1.11.0: 

```
╷
│ Error: failed to encode module.db_write_only.module.db_instance.aws_db_instance.this[0] in state: .password_wo: cannot serialize value marked as cty.NewValueMarks(marks.Ephemeral, marks.Sensitive) for inclusion in a state snapshot (this is a bug in Terraform)
│ 
│ 
│ Error: Changes.Encode: new value .password_wo: can't serialize value marked with cty.NewValueMarks(marks.Ephemeral, marks.Sensitive) (this is a bug in Terraform)
```

should be solved in 1.11.1 https://github.com/hashicorp/terraform/pull/36619

## Motivation and Context
Write Only password benefit from the fact that secrets are not stored in state anymore.

## Breaking Changes

Requires Terraform version >= 1.11

## How Has This Been Tested?

### DIFF
--- a/modules/db_instance/versions.tf
+++ b/modules/db_instance/versions.tf
@@ -1,5 +1,5 @@
 terraform {
-  required_version = ">= 1.0"
+  required_version = ">= 1.11"
 
   required_providers {
     aws = {
--- a/modules/db_instance/versions.tf
+++ b/modules/db_instance/versions.tf
@@ -1,5 +1,5 @@
 terraform {
-  required_version = ">= 1.11"
+  required_version = ">= 1.11.1"
 
   required_providers {
     aws = {
--- a/README.md
+++ b/README.md
@@ -209,7 +209,7 @@ Users have the ability to:
 
 | Name | Version |
 |------|---------|
-| <a name="requirement_terraform"></a> [terraform](#requirement\_terraform) | >= 1.11 |
+| <a name="requirement_terraform"></a> [terraform](#requirement\_terraform) | >= 1.11.1 |
 | <a name="requirement_aws"></a> [aws](#requirement\_aws) | >= 5.62 |
 
 ## Providers
--- a/README.md
+++ b/README.md
@@ -209,7 +209,7 @@ Users have the ability to:
 
 | Name | Version |
 |------|---------|
-| <a name="requirement_terraform"></a> [terraform](#requirement\_terraform) | >= 1.0 |
+| <a name="requirement_terraform"></a> [terraform](#requirement\_terraform) | >= 1.11 |
 | <a name="requirement_aws"></a> [aws](#requirement\_aws) | >= 5.62 |
 
 ## Providers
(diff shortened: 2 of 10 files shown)


modules/db_instance/versions.tf
```diff
@@ -1,5 +1,5 @@
 terraform {
-  required_version = ">= 1.0"
+  required_version = ">= 1.11.1"
```
bryantbiggs, line 2: this would be considered a breaking change, plus a large number of our users will not be on the bleeding edge

bryantbiggs:
thank you but I don't think we'll be making this change at this time

joelgaria:
@bryantbiggs Why? It seems that Terraform is following this path for security reasons, and the module should comply with these new security best practices.

bryantbiggs:
if you want to go that route, then the better practice would be to use `manage_master_user_password` 😬

joelgaria:
@bryantbiggs I believe it’s important to highlight that giving users the freedom to decide how to manage RDS passwords is crucial to ensure flexibility and versatility in workflows. For example, in some cases, it is necessary to store the password in a separate account dedicated exclusively to credential management, rather than adopting stricter solutions like manage_master_user_password.

Additionally, it is clear that the philosophy behind Terraform 1.11's [write-only arguments](https://www.hashicorp.com/en/blog/terraform-1-11-ephemeral-values-managed-resources-write-only-arguments) has extended to a broader range of resources, not just db_instance. This demonstrates that using write-only passwords is considered a significant improvement for security and the management of sensitive credentials, as highlighted in the official blog:

"Write-only arguments strengthen the security of workflows by ensuring that sensitive secrets remain ephemeral and are not exposed in Terraform's plan or state file."

Considering these aspects, keeping the issue open and evaluating the implementation of the change would allow broader and safer adoption of the credential management standards introduced in Terraform 1.11. However, enforcing the exclusive use of manage_master_user_password might discourage users who take alternative approaches tailored to their business needs.

I believe it would be better to continue the discussion around this implementation to consider the benefits it could bring to a wider range of use cases.

bryantbiggs:
> For example, in some cases, it is necessary to store the password in a separate account dedicated exclusively to credential management, rather than adopting stricter solutions like manage_master_user_password.

And how is the password provided to this module in that scenario?

bryantbiggs:
the crux of the issue is that this requires the absolute latest Terraform version and its simply not something we can just impose on users. that would be classified as a breaking change and quite impactful

edwargix:
> the crux of the issue is that this requires the absolute latest Terraform version and its simply not something we can just impose on users

Would you reconsider this PR after more time has passed for users to have upgraded to the latest terraform?  Perhaps after opentofu has considered adding the feature also (for those interested in that project)?  <https://github.com/opentofu/opentofu/issues/1996>

joelgaria:
> > For example, in some cases, it is necessary to store the password in a separate account dedicated exclusively to credential management, rather than adopting stricter solutions like manage_master_user_password.
> 
> And how is the password provided to this module in that scenario?

I run my Terraform code within a GitHub Action to securely retrieve credentials stored in a separate security account on AWS. This is made possible through the integration between GitHub and AWS. The objective is to avoid storing sensitive credentials like passwords in the Terraform state, while still deploying infrastructure that depends on them.

Here’s an example of how I achieve this:

**Secure Access to the Security Account**
By configuring the AWS Provider with an alias (aws.security), I gain access to the security account where the credentials are stored.

```
provider "aws" {
  alias  = "security"
  region = "eu-central-1"
}
```
**Using Ephemeral Resources**
I leverage the ephemeral resource so that the RDS password retrieved from AWS SSM Parameter Store is not persisted in the Terraform state file, ensuring sensitive data is managed securely.

```
ephemeral "aws_ssm_parameter" "password_rds" {
  provider = aws.security
  arn = "arn:aws:ssm:${data.aws_region.current.name}:${data.aws_caller_identity.current.account_id}:parameter/security/password_rds"
}
```
**Automatically Fetch Updates from Parameter Store**
I define a data source (aws_ssm_parameter) to fetch the password version, ensuring that updates in the Parameter Store are automatically reflected.
```
data "aws_ssm_parameter" "password_rds" {
  provider = aws.security
  name     = "/security/password_rds"
}
```
**RDS Configuration Using the Retrieved Password**
Finally, I configure the RDS module with the retrieved password. I use both the actual value and version to enhance security and enable automatic updates if the parameter changes.

```
module "test-rds" {
  source = "../modules/terraform-aws-rds"
  identifier = "test-rds"
  engine     = "mysql"
  engine_version = "8.0.40"
  username = "admin"
  password_wo         = ephemeral.aws_ssm_parameter.password_rds.value
  password_wo_version = data.aws_ssm_parameter.password_rds.version
}
```
While this example is abstract and intended for testing purposes, the approach is repeatable and adaptable to real use cases.

joelgaria:
> the crux of the issue is that this requires the absolute latest Terraform version and its simply not something we can just impose on users. that would be classified as a breaking change and quite impactful

I understand the concern about requiring the latest Terraform version and the impact it would have on users. However, sooner or later, updating Terraform will become necessary, as each new RDS feature or variable introduced will similarly require using an updated version of Terraform. For this reason, I would kindly ask you to reconsider implementing this change since the issue will inevitably arise again with future updates.

bryantbiggs:
> However, sooner or later, updating Terraform will become necessary, as each new RDS feature or variable introduced will similarly require using an updated version of Terraform.

You are confusing Terraform itself with the AWS provider here. If you look back through our module history, provider versions are routinely advanced to adopt new features of the underlying service (up to a major provider version change - this is a breaking change and occurs less frequently). Changing Terraform MSV is a less frequent change, where we recently have been updating from v1.0 to v1.3 to use variable optional attributes. Again, changing from Terraform MSV v1.0 to v1.3 is a breaking change

joelgaria:
> > However, sooner or later, updating Terraform will become necessary, as each new RDS feature or variable introduced will similarly require using an updated version of Terraform.
> 
> You are confusing Terraform itself with the AWS provider here. If you look back through our module history, provider versions are routinely advanced to adopt new features of the underlying service (up to a major provider version change - this is a breaking change and occurs less frequently). Changing Terraform MSV is a less frequent change, where we recently have been updating from v1.0 to v1.3 to use variable optional attributes. Again, changing from Terraform MSV v1.0 to v1.3 is a breaking change

I made a mistake in how I wrote my previous comment— I was actually referring not to updates to the AWS provider or RDS itself, but to how Terraform is evolving in its ability to manage RDS. For example, handling passwords without saving them in the state is an area where Terraform has made progress. I understand that this creates breaking changes, and we’ll wait for your implementation and adjust accordingly once it’s introduced. That said, this is something that will eventually need to be implemented.

antonbabenko:
This PR is included in version 7.0.0 :tada:

PR #596, opened 2025-05-09
Feature/copy db instance tags

winthropgutmann-ef:
## Description
If 'var.copy_db_instance_vars` is true then 'db_instance_tags' are merged with the common tags.

## Motivation and Context
By default the `var.datababse_instance_tags` are merged with `var.tags` which are applied to the monitoring resources. This change allows 'var.db_instance_tags' to be applied only to the rds database instance.

## Breaking Changes
No breaking changes

## How Has This Been Tested?

### DIFF
--- a/modules/db_instance/main.tf
+++ b/modules/db_instance/main.tf
@@ -137,7 +137,7 @@ resource "aws_db_instance" "this" {
     }
   }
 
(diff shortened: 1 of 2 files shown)


bryantbiggs:
thank you but this seems unnecessary and we won't be taking on this change

PR #634, opened 2026-08-27
feat: Add support for disabling the managed master user password rotation

sassdavid:
## Description

Adds `master_user_password_rotation_enabled` (`bool`, default `true`), mapped to the `rotation_enabled` argument added to `aws_secretsmanager_secret_rotation` in AWS provider v6.62.0 (hashicorp/terraform-provider-aws#49659).

Setting it to `false` cancels the rotation RDS enables by default for `manage_master_user_password` secrets:

```hcl
manage_master_user_password           = true
manage_master_user_password_rotation  = true
master_user_password_rotation_enabled = false
```

`rotation_rules` is now a `dynamic` block, since the provider rejects it when rotation is disabled.

Raises the minimum AWS provider to `>= 6.62`.

## Motivation and Context

Removes the two-apply workaround documented in `modules/db_instance/main.tf` (set `manage_master_user_password_rotation = true`, apply, set to `false`, apply again), which existed because of hashicorp/terraform-provider-aws#37779.

## Breaking Changes

No — the new variable defaults to `true`, so existing configurations plan identically. The provider floor bump is the only constraint change, happy to retitle as `feat!` if you'd rather hold this for the next major.

A future major could fold this into a tri-state `manage_master_user_password_rotation` (`null` / `true` / `false`) and drop the new variable, but that would silently cancel rotation for anyone explicitly setting `false` today, so it is not done here.

## How Has This Been Tested?

`examples/complete-postgres` now demonstrates disabling rotation on `module "db_default"` and drops the stale workaround comment on `module "db"`. Validated with `terraform validate` and `terraform plan` against aws v6.62.0 in both states (enabled → `rotation_rules` present, disabled → block omitted). not deployed against a live account.


sassdavid:
Hi @bryantbiggs, @antonbabenko, just following up on this PR as it has been open for a few weeks without a review.

If someone could take a look when you have a chance, I'd really appreciate it. Thanks!

